PR #5, opened 2026-02-01
Add support for nested meta-data inside android manifest components

Mallon94:
### Problem
- Some android services need additional meta-data within the service level of the manifest.

### Solution
 - Added buildComponentMetaData function to place in nested metadata. Can be used in future for other components in the manifest if they need nested meta-data.


simonhamp (changes requested):
Please add a test to cover this

shanerbaner82 (approved):
All 232 tests pass after merging main. Clean implementation with good test coverage.

simonhamp (approved):

PR #6, opened 2026-02-02
Added support for dictionary info_plist values

wellspringapps:
## Summary
The plist merge in the iOS compile only supports strings and arrays. There are features the require dictionary and boolean values in the plist file.

## Solution
The change was to add a method to handle all the types (String, Array, Dictionary, and Bool). 

## Changes

* src/Plugins/Compilers/IOSPluginCompiler.php - Added a method to recursively handle the structure.
* tests/Feature/Plugins/IOSCompilerTest.php - Updated `it_handles_various_plist_value_types` with all the types

### DIFF
--- a/src/Plugins/Compilers/IOSPluginCompiler.php
+++ b/src/Plugins/Compilers/IOSPluginCompiler.php
@@ -359,19 +359,8 @@ class IOSPluginCompiler
                 continue;
             }
 
-            // Handle array values
-            if (is_array($value)) {
-                $arrayContent = '';
-                foreach ($value as $item) {
-                    $item = $this->substituteEnvPlaceholders($item);
-                    $arrayContent .= "\n\t\t<string>{$item}</string>";
-                }
-                $entry = "\n\t<key>{$key}</key>\n\t<array>{$arrayContent}\n\t</array>";
-            } else {
-                // Handle string values - substitute placeholders
-                $value = $this->substituteEnvPlaceholders($value);
-                $entry = "\n\t<key>{$key}</key>\n\t<string>{$value}</string>";
-            }
+            // Handle plist types
+            $entry = $this->handlePlistTypes($key, $value);
 
             // Add before closing </dict>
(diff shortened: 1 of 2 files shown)


shanerbaner82 (changes requested):
Nice work — this fills a real gap in plist handling. A few things I'd like to see addressed before merging:

### Formatting / Output

1. **Inconsistent indentation** — nested `<dict>` and `<array>` tags all use `\n\t` regardless of depth. At 3+ levels deep the output won't match proper plist indentation. Consider passing a `$depth` parameter through the recursion to control tab count.

2. **Bool output is inconsistent** — `<true />` has a trailing `\n\t` but `<false />` doesn't. Also Apple's canonical format is `<true/>` / `<false/>` (no space before `/>`) — we should match that.

### Edge Cases

3. **Empty array handling** — `array_key_first()` returns `null` on an empty array, and `!is_int(null)` evaluates to `true`, so an empty array `[]` would be treated as a dict. Add a guard for this.

4. **Missing integer/real support** — plist also has `<integer>` and `<real>` types. If this is meant to be the comprehensive type handler, those should be covered (or at minimum, numeric values shouldn't silently become `<string>`).

### Code Quality

5. **No type hints** — `handlePlistTypes($key, $value)` is missing type declarations. The rest of the codebase uses them — this should match.

6. **`substituteEnvPlaceholders` only runs on leaf strings** — works for values, but if a placeholder ever appears in a key name it won't get substituted. Worth noting or handling.

### Tests

7. **No bool assertion** — the test sets `UIApplicationSupportsMultipleScenes => true` but never asserts that `<true/>` appears in the output. Should verify bool rendering actually works.

8. **Assertions are too shallow** — `assertStringContainsString` for a few key names doesn't validate the actual XML structure or nesting. Consider asserting on the structural tags (`<dict>`, `<array>`, `<true/>`) and/or their relative positions.

simonhamp:
Closing due to inactivity. Please feel free to reopen

PR #17, opened 2026-02-15
iOS build now respects cleanup_exclude_files config

tbtmuse:
BuildIosAppCommand::copyLaravelAppIntoIosApp() had hardcoded exclusions that ignored the user config in nativephp.cleanup_exclude_files.

Replace with the shared PlatformFileOperations::platformOptimizedCopy() already in use on the android side.

### DIFF
--- a/src/Commands/BuildIosAppCommand.php
+++ b/src/Commands/BuildIosAppCommand.php
@@ -149,7 +149,7 @@ class BuildIosAppCommand extends Command
         $source = base_path();
 
         $excludedDirs = array_merge(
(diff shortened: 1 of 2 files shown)


tbtmuse:
@simonhamp I would argue that this PR actually makes use of preexisting code rather than intruducing new code, ultimately unifying the code that is used for this function on both platforms and making it easier to reason about. It is a simplification. Would you reconsider from that perspective?

simonhamp:
This ended up being superseded by #30. Thanks

PR #18, opened 2026-02-15
Bypass WebView fetch for native event dispatch

tbtmuse:
This PR replaces the `fetch()` call with a direct platform bridge call on both Android and iOS, so native events bypass the WebView transport entirely.

## Changes

**Android** (`NativeActionCoordinator.kt`, `WebViewManager.kt`):
- Removed `fetch()` from injected JS
- Added direct `phpBridge.handleLaravelRequest()` call on a background thread
- Changed `phpBridge` visibility from `private` to `internal` in `WebViewManager`

**iOS** (`ContentView.swift`):
- Removed `fetch()` from injected JS in `notifyLaravel()`
- Added direct `NativePHPApp.laravel(request:)` call with proper `RequestData`

### DIFF
--- a/resources/androidstudio/app/src/main/java/com/nativephp/mobile/network/WebViewManager.kt
+++ b/resources/androidstudio/app/src/main/java/com/nativephp/mobile/network/WebViewManager.kt
@@ -22,7 +22,7 @@ import com.nativephp.mobile.security.LaravelSecurity
 class WebViewManager(
     private val context: Context,
     private val webView: WebView,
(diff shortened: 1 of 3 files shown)


simonhamp:
Thanks so much for the work you've put into this PR. Now that we're pushing towards fully native UI through [SuperNative](https://github.com/nativephp/mobile-air/releases/4.0.0), PRs that target the web view should focus purely on bug fixes and security patches.

PR #19, opened 2026-02-16
`native:tail` select device target and log file

Plytas:
This adds ability to select target device and log file for `native:tail` command.

Currently `tail` command allows only 1 device to be connected and reads from a single log file.  In some cases it's useful to be able to read other log files such as `browser.log` if using `laravel/boost` or daily logs if log channel is set to daily.

I extracted target device selection from `native:run` command into a `PromptsAndroidTarget` trait to reuse it.

### DIFF
--- a/src/Commands/TailCommand.php
+++ b/src/Commands/TailCommand.php
@@ -3,11 +3,20 @@
 namespace Native\Mobile\Commands;
 
 use Illuminate\Console\Command;
+use Illuminate\Support\Collection;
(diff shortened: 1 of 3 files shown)


simonhamp (changes requested):
Thanks for this. I'd really like to see this beefed up with iOS support first before we start adding more complexity here

simonhamp:
Closing due to inactivity. Please feel free to reopen

PR #23, opened 2026-02-19
feat: add Device localization support

benjam-es:
## Summary

- Add `localization()` method to `Device` returning an immutable `Localization` value object
- Exposes locale, language code, region code, timezone, currency code, and preferred language via the `Device.GetLocale` bridge function
- Add `Localization` data object under `Native\Mobile\Data` namespace
- Update `Device` facade docblock

### DIFF
--- a/config/nativephp.php
+++ b/config/nativephp.php
@@ -375,4 +375,21 @@ return [
             'landscape_right' => false,
         ],
     ],
+
+    /*
+    |--------------------------------------------------------------------------
+    | Supported Locales
+    |--------------------------------------------------------------------------
+    |
+    | The locales your app supports (e.g. ['en', 'fr', 'es', 'de']). When
+    | two or more locales are listed, iOS will resolve Locale.current using
+    | the device's preferred language (via CFBundleLocalizations in
+    | Info.plist) and Android 13+ will show a per-app language picker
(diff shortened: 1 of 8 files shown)


simonhamp:
Wouldn't `intl` support in PHP take care of this more cleanly?

benjam-es:
@simonhamp Tested this — without `CFBundleLocalizations`, the app always returned English regardless of device language. iOS scopes the locale at the process level to the bundle's development region, so `intl` would hit the same wall since it reads from the same OS source.

The build-time config tells iOS "this app supports these languages" which unblocks correct locale reporting for everything inside the app. Without it, the App Store listing also only shows the development region language, and the app won't appear in the iOS 16+ per-app language settings.

On Android, `Locale.getDefault()` works without it, but `locales_config.xml` enables the per-app language picker on 13+.

`intl` is useful for formatting once you have the locale, but the native config is needed upstream so the OS actually reports the right one.

simonhamp:
@benjam-es in the cycle towards v4, we have essentially dropped the `nativephp/mobile-device` plugin and moved its logic into the core. Would you be up for porting your work from these PRs onto v4 instead?

simonhamp:
Closing due to no activity. Feel free to reopen if this is still relevant.

PR #26, opened 2026-02-21
Auto-configure Vite plugin during native:install

benjam-es:
## Summary

- Automatically configure the `nativephpMobile` Vite plugin for Inertia (React/Vue) projects during `native:install`
- Inject import, `hotFile` config, and `nativephpMobile()` into the user's vite config
- Validate the modified config with `node --check` before overwriting
- Fall back to manual instructions if auto-config can't be applied or user declines
- Add `--with-vite` flag to skip the confirmation prompt

## Test plan

- [x] 10 unit tests covering: skip conditions, injection, idempotency, decline path, temp file cleanup
- [x] PHPStan passes
- [x] Pint passes

## Related

- Docs PR: https://github.com/NativePHP/nativephp.com/pull/270

### DIFF
--- a/src/Commands/InstallCommand.php
+++ b/src/Commands/InstallCommand.php
@@ -8,12 +8,14 @@ use Native\Mobile\Traits\InstallsAndroid;
 use Native\Mobile\Traits\InstallsIos;
 use Native\Mobile\Traits\PlatformFileOperations;
 
+use function Laravel\Prompts\confirm;
 use function Laravel\Prompts\error;
 use function Laravel\Prompts\intro;
 use function Laravel\Prompts\note;
 use function Laravel\Prompts\outro;
 use function Laravel\Prompts\select;
 use function Laravel\Prompts\text;
+use function Laravel\Prompts\warning;
 
 class InstallCommand extends Command
 {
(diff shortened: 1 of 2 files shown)


simonhamp (changes requested):
I love the idea of this and it was something I was exploring when creating the Vite plugins, but using regex to modify the Vite config is super fragile

The better approach (which I didn't have time for) is a to use an AST to modify the existing code

But a simpler short-term solution is to just write a new pre-configured vite.config.js file

Or better yet build an app from our new starter kit in which it is [already pre-configured](https://github.com/NativePHP/mobile-starter/blob/main/vite.config.js)

benjam-es (commented):
Totally fair point on regex fragility in general — but I think the Laravel Vite config is a special case. It's essentially a template: short, predictable structure, and users rarely deviate much from what the starter gives them.

The current approach also has safety nets: it writes to a temp file, validates with `node --check` before overwriting, and falls back to manual instructions if anything goes wrong. So worst case, the user just gets told what to paste.

The AST route would mean either bundling a parser or adding a dependency — feels like a lot of overhead for modifying what's effectively a 15-line file.

Happy to explore writing a fresh config file instead if you'd prefer that, but wanted to make the case for keeping it simple.

Thoughts @simonhamp ?

simonhamp:
I don't know that this is simple. I know a lot of projects that have all sorts of quirky logic in their Vite config that this probably wouldn't work with.

I think the changes needed to the config is well covered in the docs.

Thanks for your effort though!

PR #30, opened 2026-02-22
feat(ios): replace PHP iterator with rsync for app copy step

benjam-es:
## Problem

`BuildIosAppCommand::copyLaravelAppIntoIosApp()` uses a `RecursiveIteratorIterator` with `FOLLOW_SYMLINKS` to copy the Laravel app into the iOS build directory. This approach has three issues:

1. **Memory** — Every file is collected into a PHP array before copying. On a standard Inertia project this consumes ~60 MB of memory. With local Composer path repositories (symlinked packages), the iterator follows symlinks into nested `vendor/` directories, easily exceeding the default 128 MB memory limit and causing a silent exit code 255 with no error output.

2. **Wasted I/O** — The iterator copies files that `removeUnnecessaryFiles()` immediately deletes afterwards (notably `node_modules`). On a real project this means ~19,000 files and ~240 MB of unnecessary disk writes every build.

3. **Speed** — The two-pass approach (collect all files into memory, then iterate and copy) is slower than a native file copy tool.

## Solution

Replace the PHP iterator with a single `rsync -a --copy-links` call using exclude patterns. A new `getExcludedPaths()` method provides a single source of truth for paths that should never enter the iOS bundle. Vendor package `.gitattributes` `export-ignore` patterns are also respected when present (relevant for locally symlinked packages during development).

`removeUnnecessaryFiles()` is unchanged — it continues to clean up after `composer install` which may recreate excluded directories.

## Benchmarks

Profiled on **kitchen-sink-mobile-vue** (Vue/Inertia, 14 NativePHP plugins, stock mobile ^3.0, no symlinks):

| | **Before (iterator)** | **After (rsync)** |
|---|---|---|
| Files iterated by PHP | 36,116 | 0 |
| Files copied | 30,865 | 10,484 |
| Files immediately deleted by cleanup | 19,032 | 13 |
| Wasted I/O | 240 MB | 0 |
| Copy time | 7,702 ms | 2,500 ms |
| Cleanup time | 1,418 ms | 17 ms |
| **Total time (copy + composer + cleanup)** | **11,250 ms** | **3,291 ms** |
| **Peak PHP memory** | **60.5 MB** | **2 MB** |
| Final bundle files | 11,833 | 10,471 |
| Final bundle size | 102 MB | 93 MB |

With **locally symlinked packages** (3 Composer path repos), the original iterator follows into each package's own `node_modules/` and `vendor/` directories:

| | **Before** | **After** |
|---|---|---|
| Files iterated | 59,741 | 0 |
| Peak PHP memory | 104+ MB (OOM at 128 MB) | 2 MB |
| Build result | **Silent crash (exit 255)** | **Completes normally** |

## What changed

- `copyLaravelAppIntoIosApp()` — replaced `RecursiveIteratorIterator` + manual copy loop with `rsync -a --copy-links` and exclude flags
- `getExcludedPaths()` — new method, single source of truth for excluded paths (used by rsync)
- `loadVendorExportIgnorePatterns()` — new method, reads `.gitattributes` `export-ignore` from vendor packages
- `removeUnnecessaryFiles()` — added glob support for wildcard patterns in `$directoriesToRemove`, added `vendor/*/*/vendor` cleanup as safety net
- Progress feedback — copy step wrapped in `$this->components->task()`

## Test plan

- [x] Profiled on kitchen-sink-mobile-vue (stock mobile ^3.0, no symlinks)
- [x] Profiled on fresh React Inertia app with all plugins
- [x] Tested with locally symlinked Composer path repositories (3 packages)
- [x] Verified final bundle contents match expected output
- [x] Confirmed `removeUnnecessaryFiles()` still cleans up after `composer install`


simonhamp (approved):
I've tested this out on Mac and it's all working. It does feel a bit faster and it's done everything I expected it to do. Great work!

PR #76, opened 2026-03-30
Fix package command returning exit code 0 on build failure

shanerbaner82:
## Summary
- `handle()` returned `void`, so `native:package` always exited with code 0 — even when Gradle failed or no output file was produced. CI (Bifrost) had no way to detect build failures via exit code.
- Gradle output was swallowed in `--no-tty` mode because `Process::run()` captured stdout but nothing printed it — builds appeared to hang then silently succeed.
- `handleBuildArtifacts()` returning `null` (no AAB/APK found) was never checked — the command printed "Signed bundle build complete" regardless.

## Changes
- `handle()` returns `int` (1 on error, 0 on success)
- `buildAndroid()` returns `bool` so `handle()` can propagate failure
- Added check: if `handleBuildArtifacts()` returns null, error and return false
- Added output callback to `Process::run()` so Gradle output streams in real-time during `--no-tty` builds

## Test plan
- [ ] Run `php artisan native:package android --build-type=bundle --no-tty` — confirm Gradle output streams to terminal
- [ ] Break build.gradle.kts and run package — confirm command exits with code 1
- [ ] Run successful build — confirm exit code 0, AAB produced, output shown
- [ ] Trigger Bifrost CI build — confirm failed builds now report as failed

### DIFF
--- a/src/Commands/PackageCommand.php
+++ b/src/Commands/PackageCommand.php
@@ -63,7 +63,7 @@ class PackageCommand extends Command
 
     protected string $platform;
 
-    public function handle(): void
+    public function handle(): int
     {
         // Get platform (flags take priority over argument)
         if ($this->option('ios')) {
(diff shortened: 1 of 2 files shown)


shanerbaner82:
Stale

PR #77, opened 2026-03-30
Fix iOS top bar breaking edge-to-edge webview layout

shanerbaner82:
## Summary
- Changed top bar from `safeAreaInset(edge: .top)` to a ZStack overlay so the webview stays edge-to-edge behind the status bar and top bar
- Updated `--inset-top` CSS injection to include nav bar height (52pt) when top bar is present, so `.nativephp-safe-area` padding still works
- Removed top-bar-dependent `ignoresSafeArea` conditions — webview now always ignores all safe areas

## Test plan
- [ ] Verify red/colored backgrounds extend edge-to-edge behind status bar with top bar present
- [ ] Verify content text has padding below the top bar (not flush against it)
- [ ] Verify layout still works correctly without a top bar
- [ ] Verify bottom nav layouts (iOS 26+ and iOS 18) are unaffected
- [ ] Test with side nav — menu button should still open sidebar

### DIFF
--- a/resources/xcode/NativePHP/ContentView.swift
+++ b/resources/xcode/NativePHP/ContentView.swift
@@ -12,14 +12,19 @@ struct ContentView: View {
     @StateObject private var uiState = NativeUIState.shared
 
     var body: some View {
-        NativeSideNavigation(onNavigate: handleNavigation) {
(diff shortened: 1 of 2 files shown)


simonhamp:
Do we still need this?

shanerbaner82:
Stale

PR #86, opened 2026-04-07
Fix plugin directory structure inconsistency in PluginCreateCommand

wilsonatb:
*Updated PR based on [simonhamp](https://github.com/simonhamp) feedback to adapt the validator instead of the creator.*

### Problem Description
Freshly created plugins generate a flat directory structure (`resources/android/` and `resources/ios/`), but `PluginValidateCommand` strictly expects a nested structure (`/src/` and `/Sources/`), resulting in false positive warnings like `No native code directories found`.

### Solution
Instead of forcing the creation command to use nested folders, this PR updates `PluginValidateCommand` to support both architectures. 

**What changed in `PluginValidateCommand.php`:**
1. The validator now checks for the nested structure first.
2. If not found, it gracefully falls back to checking the flat structure.
3. It only emits warnings if the implementation is missing in *both* locations.
This aligns the validator with the core `Plugin` class which already supports both approaches seamlessly.

### Testing
- [x] Tested locally by generating a fresh plugin with flat structure.
- [x] Ran `native:plugin:validate` (passed with a clean `OK` and 0 warnings).

### Visual Proof (Before & After)

**Before (Current Main):**
```text
$ php artisan native:plugin:validate packages/wilsonatb/plugin-test-bug
  plugin-test-bug (android: 21, ios: 15.0) ...................................................................... WARN  
  ⚠ No native code directories found (resources/android or resources/ios)
  ⚠ Android implementation not found for TestBug.Execute
  ⚠ iOS implementation not found for TestBug.Execute
```

**After (This PR):**
```text
$ php artisan native:plugin:validate packages/wilsonatb/plugin-test-fallback
  plugin-test-fallback (android: 21, ios: 15.0) ........................................................................ OK 
```

### DIFF
--- a/src/Commands/PluginValidateCommand.php
+++ b/src/Commands/PluginValidateCommand.php
@@ -201,8 +201,10 @@ class PluginValidateCommand extends Command
         ];
 
         $optionalDirs = [
-            'resources/android/src' => 'Android Kotlin sources',
-            'resources/ios/Sources' => 'iOS Swift sources',
+            'resources/android/src' => 'Android Kotlin sources (nested)',
+            'resources/android' => 'Android Kotlin sources (flat)',
+            'resources/ios/Sources' => 'iOS Swift sources (nested)',
+            'resources/ios' => 'iOS Swift sources (flat)',
         ];
 
         foreach ($requiredDirs as $dir => $description) {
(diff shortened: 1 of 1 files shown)


simonhamp:
I think I'd probably prefer this the other way around: so the validator changes

Thoughts @shanerbaner82 ?

wilsonatb:
@simonhamp  I completely agree with your approach. It makes much more sense to make the validator smarter rather than forcing a specific scaffolding structure, especially since the `Plugin` class already handles both gracefully.

I've updated the PR. I reverted the changes to the creator and instead updated `PluginValidateCommand` to fallback to the flat structure if the nested one is not found. Tested locally and works perfectly!

simonhamp (approved):

PR #93, opened 2026-04-13
fix(ios): preserve URL percent-encoding through to PHP

bkuhl:
## Problem

Laravel's URL generator (`route()`, `redirect()`, etc.) percent-encodes path and query values. PHP expects to consume the URI with that encoding intact — the shape `$_SERVER['REQUEST_URI']` and `$_SERVER['QUERY_STRING']` would have behind a real HTTP server.

**Android** preserves the encoding via `Uri.encodedPath` (added in #24).

**iOS** does not. The current handler uses `URL.path`, which percent-decodes once before handing the URI to PHP:

```swift
let uri = request.url?.path ?? "/"
```

When Laravel generates a URL like `/game-board/abc%2Fdef`, PHP on iOS sees `/game-board/abc/def`. Reserved characters (`/`, `+`, `$`, `*`) and literal `%` from route parameters arrive as bare characters, breaking route matching — Symfony rejects URIs Laravel itself generated as malformed (HTTP 400).

The encoding Laravel produces and the encoding iOS NativePHP delivers do not match.

## Fix

Use `URLComponents.percentEncodedPath` and `.percentEncodedQuery` — the iOS-equivalent of Android's `Uri.encodedPath`. PHP receives the URI in the same shape on both platforms, matching what Laravel and Symfony expect.

This PR also depends on https://github.com/NativePHP/mobile-air/pull/108 being merged, as 108 actually fails first before it reaches this point.

## Test plan

- [x] Routes with reserved chars in path (`/`, `+`, `$`, `*`)
- [x] Routes with literal `%` in route parameter data
- [x] Long URLs (~3500 char real-world tokens)
- [x] Static asset routes
- [x] Routes without special characters

### DIFF
--- a/resources/xcode/NativePHP/PHPSchemeHandler.swift
+++ b/resources/xcode/NativePHP/PHPSchemeHandler.swift
@@ -517,57 +517,11 @@ class PHPSchemeHandler: NSObject, WKURLSchemeHandler {
 
                 var request = requestData
                 if let location = headers["location"] {
-                    let trimmedLocation = location.trimmingCharacters(in: .whitespaces)
+                    // Location headers come directly from PHP's HTTP response — they are
+                    // raw strings that never pass through WebKit's URL normalization, so
+                    // they do NOT need re-encoding (unlike initial requests in extractRequestData).
+                    request.uri = location.trimmingCharacters(in: .whitespaces)
                     request.method = "GET"
-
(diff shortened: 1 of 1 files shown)


simonhamp (approved):

simonhamp:
Thanks Ben!

bkuhl:
Thank you!  I'm deleting my feedback in Discord - appreciate you giving this a review!

PR #104, opened 2026-04-25
fix: pin to correct PHP version

1338:
- Set of packages are currently only installable on PHPv8.4


1338:
I can no longer be active in Open Source and therefor must close this PR/ticket.

PR #112, opened 2026-05-06
fix(ios): avoid format-string crash in PHPSchemeHandler logs

shreejanpandit:
## What does this PR do?
- Replaces two `NSLog` calls in `PHPSchemeHandler.getResponse()` with `print` for request/response tracing.
- Prevents native iOS crashes caused when response/request strings include `%` sequences that `NSLog` interprets as format specifiers.
- Keeps logging visibility while avoiding format-string parsing on the persistent PHP queue.

## Risks & Mitigation
- Low risk: change is limited to debug logging statements and does not alter request dispatch or response/cookie behavior.
- Mitigation: only two lines are changed in one file, with no control-flow or data-path changes.
- If needed, behavior can be reverted safely without affecting runtime protocol handling.

## How was this tested?
- Reproduced crash locally from app actions that route through `PHPSchemeHandler`.
- Applied this exact change in the app runtime copy and confirmed simulator + physical device no longer crashed on the same actions.
- Verified the same patch is now committed on this fork branch for upstream review.

Made with [Cursor](https://cursor.com)


simonhamp:
This was already fixed in #108

PR #126, opened 2026-05-11
Fix native:jump hang on Windows and harden port detection

simonhamp:
## Summary

Fixes a report from a Windows user that `php artisan native:jump` got stuck after printing the Laravel server line, never rendering the QR code.

Three related fixes in `src/Commands/JumpCommand.php`:

- **Drop the `killExistingServers` cleanup step.** With reliable dynamic port selection in place, we don't need to hunt and kill prior processes — and the cleanup never reliably caught stray `artisan serve` instances anyway. Also removes the noisy "Cleaning up N existing server(s)" line.

- **Harden `isPortInUse()` with a bind-test alongside the existing connect-test.** `fsockopen` only detects ports that are *accepting* connections; it misses ports held but not listening (stuck/half-dead Windows servers, `TIME_WAIT`, bound-but-not-listening). That's why a stale previous `artisan serve` on 8000 could fool the old check into selecting 8000 again. We now also try `stream_socket_server` and treat a bind failure as in-use.

- **Fix the actual Windows hang in `startBridgeServer()`.** The bridge spawn used a trailing `&` to background the websocket server, which is Unix shell syntax. On Windows, `&` is a command separator — not a background operator — so `exec()` blocked forever on the long-lived bridge server, which is why the QR code never appeared. Windows now uses `start "" /B ... >> log 2>&1` via `popen`/`pclose` for a proper detached spawn; Unix path is unchanged.

## Test plan

- [ ] On macOS/Linux: `php artisan native:jump` still starts, picks ports, shows QR, and the bridge log fills as the phone connects.
- [ ] On Windows: `php artisan native:jump` reaches the QR display without hanging, even when a prior `artisan serve` is still running on 8000.
- [ ] Re-run `php artisan native:jump` repeatedly without manually killing prior servers — each run picks fresh ports and starts cleanly.

### DIFF
--- a/src/Commands/JumpCommand.php
+++ b/src/Commands/JumpCommand.php
@@ -39,9 +39,6 @@ class JumpCommand extends Command
 
         intro('NativePHP Jump Server');
 
-        // Kill existing servers
-        $this->killExistingServers();
-
         // Configuration
         $host = $this->option('host');
         $httpPort = $this->option('http-port') ?? config('nativephp.server.http_port', 3000);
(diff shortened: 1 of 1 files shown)


ZeoNish:
Win 10 pro x64 
nativephp 3.3.4 (and 3.3.3, 3.3.2 ...)

image
this console not close

next start more new console ConEmu
image

start qr

but console phpstorm
image

qr to console

ZeoNish:
The patch solves the problem

image

ZeoNish:
wow. delete "nativephp/mobile" composer and reinstall 
image

solves the problem too O_O ... wtf...

mustafa-online:
This fixed the QR problem 🥇 
It's now showing on Terminal / Windows 11

PR #131, opened 2026-05-14
Make Android theme colors configurable

### LINKED ISSUE #42: Dark mode issues on Android

There seems to be an issue with a time picker on Android for people using dark mode. The buttons below the clock are not visible at all.

In another Andoid app (not built with NativePHP) this issue doesn't seem to happen. 

Is this something we can specify somewhere?

Image

Image

### What happened?

The buttons below the clock are not visible at all.

### How to reproduce the bug

Set dark mode on android and open a time picker 

### Debug Output

none

_No response_

Android (Emulator), Android (Device)

### Notes

The buttons below the clock are not visible at all.

voicecode-bv:
This update allows users to choose their own colors for the Android theme. This makes it possible for date and time pickers to be displayed in the app's branding and resolves the issue where buttons were not visible in dark mode.

Fixes #42

### DIFF
--- a/config/nativephp.php
+++ b/config/nativephp.php
@@ -225,9 +225,9 @@ return [
         |
         */
         'theme' => [
-            'color_primary' => env('NATIVEPHP_ANDROID_COLOR_PRIMARY') ?: '#04ABA6',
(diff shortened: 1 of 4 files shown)


simonhamp (approved):
Looks great! Thanks

PR #135, opened 2026-05-18
Fix native:jump hang after IP select on Windows

simonhamp:
## Summary

Windows users running `php artisan native:jump` saw the TUI freeze immediately after picking an IP from the "Multiple network interfaces detected" prompt — the QR code never rendered.

The previous fix (#126) switched bridge-server spawning to `pclose(popen("start \"\" /B ... >> log 2>&1", "r"))`. That still hangs on some setups: `popen` inherits the cmd.exe stdout pipe, `start /B` propagates that handle into the long-lived bridge grandchild via `CreateProcess(bInheritHandles=TRUE)`, and combined with append redirection the pclose call ends up blocked, so execution never reaches the QR render or any of the subsequent `twoColumnDetail` lines.

## Fix

Spawn the bridge via `proc_open` with `bypass_shell => true` and explicit file descriptors:

- stdin → `NUL`
- stdout / stderr → the log file

No cmd.exe sits in the middle, so no inheritable pipe is ever created, and `proc_open` returns immediately after `CreateProcess` succeeds. The resource is held on `$this->bridgeProcess` so its blocking destructor doesn't run mid-command. On Ctrl+C, Windows hard-terminates the PHP CLI without running destructors, so the bridge process is left orphaned — matching the existing Mac/Linux behaviour where `exec($cmd . ' &')` also doesn't clean up.

Non-Windows path is unchanged.

## Test plan

- [ ] On Windows: `php artisan native:jump` with multiple network interfaces — confirm the IP-select prompt no longer freezes and the QR renders.
- [ ] On Windows: confirm `storage/logs/jump-bridge.log` shows the bridge server starting and accepting connections.
- [ ] On Windows: confirm Ctrl+C exits cleanly (bridge may remain running — matches Mac/Linux).
- [ ] On macOS / Linux: confirm `native:jump` still works (path is unchanged).

### DIFF
--- a/src/Commands/JumpCommand.php
+++ b/src/Commands/JumpCommand.php
@@ -31,6 +31,8 @@ class JumpCommand extends Command
 
     private array $laravelPipes = [];
 
+    private $bridgeProcess = null;
+
     private bool $verbose = false;
 
     public function handle()
@@ -299,21 +301,37 @@ class JumpCommand extends Command
         @file_put_contents($logFile, '=== '.date('Y-m-d H:i:s')." bridge server starting (ws={$wsPort} tcp={$bridgePort} vite_proxy={$viteProxyPort}) ===\n", FILE_APPEND);
 
         // Run in background (not Workerman daemon mode — it breaks the event loop).
-        // Windows: `&` is a command separator (not "background"), so exec() would
-        // block here forever waiting on the long-lived server. Use `start /B` to
-        // detach properly.
         if (PHP_OS_FAMILY === 'Windows') {
+            // `&` is a command separator on Windows (not "background"), and the
+            // previous `pclose(popen("start /B ..."))` approach hangs: cmd.exe's
+            // stdout pipe (created by popen) gets inherited by the grandchild
+            // PHP via CreateProcess(bInheritHandles=TRUE) and the pipe never
+            // sees EOF, so pclose blocks until the long-lived bridge exits.
+            //
+            // Use proc_open with explicit file handles (no inheritable pipes)
+            // and bypass_shell so no cmd.exe sits in the middle. Intentionally
+            // do NOT proc_close the resource — that would wait on the
+            // long-running child. The OS process is independent and the PHP
+            // resource is cleaned up at script shutdown.
             $cmd = sprintf(
-                'start "" /B %s %s %s %d %d %d start >> %s 2>&1',
+                '%s %s %s %d %d %d start',
                 escapeshellarg($phpBinary),
                 escapeshellarg($serverPath),
                 escapeshellarg(base_path()),
                 $wsPort,
                 $bridgePort,
-                $viteProxyPort,
-                escapeshellarg($logFile)
+                $viteProxyPort
             );
-            pclose(popen($cmd, 'r'));
+            $desc = [
+                0 => ['file', 'NUL', 'r'],
+                1 => ['file', $logFile, 'a'],
+                2 => ['file', $logFile, 'a'],
+            ];
+            // Keep the resource on the instance so its destructor doesn't fire
+            // mid-command (proc_close blocks waiting for the long-lived child).
+            // On Ctrl+C the PHP process is hard-terminated by Windows and the
+            // bridge stays running, matching Mac/Linux behaviour.
+            $this->bridgeProcess = @proc_open($cmd, $desc, $pipes, base_path(), null, ['bypass_shell' => true]);
         } else {
             $cmd = sprintf(
                 '%s %s %s %d %d %d start >> %s 2>&1 &',


PalalauAlexandru:
Confirmed this PR fixes the Windows hang after selecting an IP in my setup.

Environment:
- Windows
- Laravel Herd
- PHP 8.5 via Herd
- NativePHP Jump
- Multiple network interfaces

After applying this PR locally, `php artisan native:jump` no longer freezes after the IP selection. The QR code renders, the mobile app can scan it, and the device connects to the Jump bridge successfully.

However, this uncovered the next Windows/Herd-specific blocker:

```txt
Bad Gateway: Could not connect to Laravel on port 8000.
Error: Failed to connect to 127.0.0.1 port 8000
```

Diagnosis:

Jump itself was running correctly:

```txt
0.0.0.0:3000       LISTENING   # Jump router
0.0.0.0:3001       LISTENING   # WebSocket bridge
127.0.0.1:3002     LISTENING   # TCP bridge
```

The phone was connected to the bridge, but nothing was listening on `127.0.0.1:8000`.

Running Laravel's managed serve command manually fails in this Herd/Windows setup:

```bash
php artisan serve --host=127.0.0.1 --port=8000 --no-interaction
```

Output:

```txt
Failed to listen on 127.0.0.1:8000 (reason: ?)
```

But starting PHP's built-in server directly works:

```bash
php -S 127.0.0.1:8000 vendor/laravel/framework/src/Illuminate/Foundation/resources/server.php
```

After switching Jump's Windows Laravel-server startup path to direct `php -S ... server.php`, the Jump proxy started responding correctly:

```txt
http://127.0.0.1:8000/ -> 200
http://127.0.0.1:3000/ -> 200
```

This may be separate from this PR, but it appears immediately after the QR/bridge fix succeeds, so I wanted to document it here.

Possible follow-up patch, Windows-only:

```diff
diff --git a/src/Commands/JumpCommand.php b/src/Commands/JumpCommand.php
--- a/src/Commands/JumpCommand.php
+++ b/src/Commands/JumpCommand.php
@@ -364,14 +364,28 @@ private function startLaravelServer(int $port, int $bridgePort = 3002, int $wsPo
         $phpBinary = PHP_BINARY;
         $artisan = base_path('artisan');

         $descriptorSpec = [
             0 => ['pipe', 'r'],
             1 => ['pipe', 'w'],
             2 => ['pipe', 'w'],
         ];

-        $cmd = sprintf(
-            '%s %s serve --port=%d --host=127.0.0.1 --no-interaction',
-            escapeshellarg($phpBinary),
-            escapeshellarg($artisan),
-            $port
-        );
+        if (PHP_OS_FAMILY === 'Windows') {
+            $server = file_exists(base_path('server.php'))
+                ? base_path('server.php')
+                : base_path('vendor/laravel/framework/src/Illuminate/Foundation/resources/server.php');
+
+            $cmd = sprintf(
+                '%s -S 127.0.0.1:%d %s',
+                escapeshellarg($phpBinary),
+                $port,
+                escapeshellarg($server)
+            );
+        } else {
+            $cmd = sprintf(
+                '%s %s serve --port=%d --host=127.0.0.1 --no-interaction',
+                escapeshellarg($phpBinary),
+                escapeshellarg($artisan),
+                $port
+            );
+        }

         // Pass bridge ports so nativephp_call() (JumpBridge) in Laravel dials the right TCP port.
         $env = array_merge($_ENV, $_SERVER, [
             'JUMP_BRIDGE_PORT' => (string) $bridgePort,
             'JUMP_WS_PORT' => (string) $wsPort,
         ]);
         $env = array_filter($env, fn ($v) => is_string($v) || is_numeric($v));

-        $this->laravelProcess = proc_open($cmd, $descriptorSpec, $this->laravelPipes, base_path(), $env);
+        $this->laravelProcess = proc_open(
+            $cmd,
+            $descriptorSpec,
+            $this->laravelPipes,
+            PHP_OS_FAMILY === 'Windows' ? public_path() : base_path(),
+            $env
+        );
```

This keeps the existing non-Windows path unchanged and only bypasses `artisan serve` on Windows, where Herd / Symfony Process / Laravel serve can fail to bind even though direct `php -S` works.

Additional observation after more testing:

After the Windows/Herd workaround above, Jump can start successfully and the phone can browse the app, but I sometimes hit this later in the Android WebView:

```txt
Page Web non disponible
net::ERR_CONNECTION_RESET on http://<LAN-IP>:3000/
```

Once that happens, the app does not recover until `php artisan native:jump` is restarted.

In this state, Laravel itself may still be healthy on the internal Laravel port, but the Jump HTTP router on port `3000` can stop responding or disappear. Example local checks showed `127.0.0.1:<laravel-port>` responding while `127.0.0.1:3000/jump/info` no longer responded.

A local mitigation I am testing is a small Jump-router watchdog in `JumpCommand`:

- Start the Jump PHP built-in server with `public/` as the docroot:
  `php -S 0.0.0.0:3000 -t public router.php`
- Periodically call `GET /jump/info` on the local Jump router.
- If the Jump router process exits, restart it.
- If the Jump router process is still running but stops answering HTTP, restart it.
- Keep the existing Laravel-server watchdog separately, because the Laravel server and the Jump router can fail independently.

I also tested a router-side mitigation:

- Do not proxy `/jump/download` to Laravel; return a direct Jump-side `404` instead.
- Use fresh/non-reused cURL connections when proxying to Laravel.
- Use shorter timeouts for static reads such as `/build`, `/images`, and `/storage`, because the PHP built-in server is single-process and a slow static request can block the Jump router for too long.

This part may be a separate Windows/Herd stability issue rather than part of this PR. I would not necessarily include the watchdog directly in this PR, but it may be useful context for a follow-up issue after the QR/IP-selection hang is fixed.

simonhamp:
@PalalauAlexandru I'm closing this in favour of #136. Could you give that branch a try and let me know how you get on. Sorry for jumping this around

PR #136, opened 2026-05-18
Potential jump win fix

shanerbaner82:
All changes scoped to nativephp/mobile (Herd\mobile-air). macOS/Linux code paths untouched (PHP_OS_FAMILY === 'Windows' gates).

src/Commands/JumpCommand.php

warmLaravelServer(): curls GET / after artisan serve boots so the first phone request hits warm Laravel (cold boot was exceeding the proxy's curl timeout → opaque 502 in the phone reading "Could not connect to Laravel on port 8000").
startPhpServer is platform-gated: Windows spawns php http-server.php; Mac/Linux still spawns php -S … router.php.
startViteHmrProxyForWindows(): additionally spawns vite-hmr-server.php because Workerman silently drops the 2nd Worker declared in websocket-server.php on Windows (no fork), which is why HMR never bound to 3003.
resources/jump/router.php (Mac/Linux only now, but improvements apply both)

Diagnostic 502 bodies (CURLE_COULDNT_CONNECT vs CURLE_OPERATION_TIMEDOUT vs other), [start] + completion logs, proxyToVite timeout 10→30s, proxyToLaravel 30→90s, Connection: close to drain keepalive sockets, request logs split to storage/logs/jump-router.log with ms precision.
New: resources/jump/http-server.php — Workerman HTTP worker, feature-for-feature port of router.php. Closes TCP after each response, eliminating the dead-socket wedge php -S on Windows hits because it doesn't set SO_KEEPALIVE.

New: resources/jump/vite-hmr-server.php — standalone copy of the JumpViteProxy Worker from websocket-server.php so it can run as its own Windows process.

### DIFF
--- a/config/nativephp.php
+++ b/config/nativephp.php
@@ -265,8 +265,10 @@ return [
         // Build output directory (where the ZIP will be created)
         'build_path' => env('NATIVEPHP_BUILD_PATH', 'storage/app/native-build'),
 
-        // Automatically open browser with QR code when server starts
-        'open_browser' => env('NATIVEPHP_OPEN_BROWSER', true),
+        // Automatically open browser with QR code when server starts.
+        // Default off — the terminal renders a scannable QR. Pass --browser
+        // to native:jump (or set NATIVEPHP_OPEN_BROWSER=true) to opt in.
+        'open_browser' => env('NATIVEPHP_OPEN_BROWSER', false),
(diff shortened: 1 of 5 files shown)


simonhamp:
supersedes #135

simonhamp (approved):

Issue #139, opened 2026-05-19
iOS: WKWebsiteDataStore.nonPersistent() wipes cookies on every cold launch — Laravel session lost


sawan-webmavens:
### Summary

The iOS shell uses `WKWebsiteDataStore.nonPersistent()` in `resources/xcode/NativePHP/ContentView.swift:149`. This is in-memory storage, so the WKWebView starts with an empty cookie jar on every cold launch. For Laravel apps this means the `laravel_session` cookie is lost — the session file is still on disk in `storage/framework/sessions/`, but Laravel can't find it without the cookie that identifies it. The user appears signed-out, and all session-keyed state (auth user, flash messages, dismissed prompts, anything stored via `session(...)`) resets.

Android uses `CookieManager.getInstance()` (`app/src/main/java/com/nativephp/mobile/network/WebViewManager.kt`, in `setupCookieManager()`) — persistent and shared by default. So iOS and Android behave inconsistently for the same Laravel app: Android remembers your session across cold launches, iOS does not.

### Reproduction

Any NativePHP iOS app that signs a user in via standard Laravel auth, then cold-launches. On the next launch, `auth()->check()` is `false` despite no logout occurring. Anything stored under `session(...)` is gone. Auth tokens stored in `SecureStorage` (Keychain) survive, which can mask the bug — the app continues making authenticated API calls while Laravel itself sees the user as a guest.

### Proposed fix

Switch to `WKWebsiteDataStore.default()` — the standard persistent jar. One-line change. Happy to PR.

### Open design question

Should this be:

1. **An unconditional flip** — matches Android, fixes the bug for everyone, no config surface area.
2. **A config option** — e.g. `'cookies' => 'persistent'|'ephemeral'` in `config/nativephp.php`, defaulting to `'persistent'`. Preserves the current behavior for anyone who actually wanted it.

My read is unconditional, since the current behavior is almost certainly never what anyone wants for a Laravel-backed app, but I'll go with whatever you prefer.

sawan-webmavens:
PR up: #140 (unconditional flip — happy to switch to a config option if maintainers prefer).

simonhamp:
#140 got merged and released

PR #140, opened 2026-05-19
fix(ios): use persistent WKWebsiteDataStore so cookies survive cold launch

sawan-webmavens:
## Summary

- Switch iOS WKWebView from `WKWebsiteDataStore.nonPersistent()` to `.default()` so cookies persist across cold launches.
- Aligns iOS behavior with Android, which already uses persistent storage via `CookieManager.getInstance()`.
- Fixes the Laravel session loss documented in #139.

## Why

`WKWebsiteDataStore.nonPersistent()` is in-memory storage — the WKWebView starts with an empty cookie jar on every cold launch. For Laravel apps this wipes the `laravel_session` cookie, so even though the session file is still on disk in `storage/framework/sessions/`, Laravel can't find it. The user appears signed-out, and any `session(...)` state (auth user, flash messages, dismissed prompts) is reset.

Android already does the right thing — `CookieManager.getInstance()` is persistent and shared by default. Same Laravel code, two different behaviors. This patch brings iOS in line.

## Risk

`WKWebsiteDataStore.default()` is the shared, app-wide persistent jar. Not an issue for NativePHP apps (they own their entire WebView), but flagging in case reviewers want it on the radar.

## Test plan

- [ ] Cold-launch a NativePHP iOS app that signs the user in via Laravel auth. Confirm `auth()->check()` is `true` on next cold launch (was `false` before this change).
- [ ] Confirm `session(...)` state survives cold launch.
- [ ] Verify behavior matches Android same-app behavior.

## Notes

Refs #139, which raised the open question of unconditional flip vs. config option. This PR is the unconditional flip — happy to amend to a config option (e.g. `'cookies' => 'persistent'|'ephemeral'` defaulting to persistent) if you'd rather preserve the current behavior as an opt-in.

### DIFF
--- a/resources/xcode/NativePHP/ContentView.swift
+++ b/resources/xcode/NativePHP/ContentView.swift
@@ -146,7 +146,7 @@ struct WebViewLayoutContainer: View {
 }
 
 struct WebView: UIViewRepresentable {
-    static let dataStore = WKWebsiteDataStore.nonPersistent()
+    static let dataStore = WKWebsiteDataStore.default()
     let shared: SharedWebView
     let horizontalSizeClass: UserInterfaceSizeClass?
 


simonhamp:
I think it's right to have consistent behaviour across both platforms 👍🏼 

My caveat is that I think cookies as a mechanism for determining authenticated state is kind of "wrong" (convenient, but not as secure as secure storage) in this context and the fact that Android supports it actually suggests, imo, that it's Android that should change to align with iOS.

Just my opinion tho and some history on why I chose for iOS cookie jars to be in-memory only in the first place.

simonhamp:
I've gone back and forth on this a bit and I'm happy for this change to come in as-is. Especially as we're moving away from the web view. Merging

PR #142, opened 2026-05-23
Fix iOS LaunchImage Contents.json mixing Any-scale + specific scales

### LINKED ISSUE #141: iOS LaunchImage Contents.json mixes 'Any' scale with specific scales when all 6 splash variants exist (asset compile fails)
## Summary

`InstallsSplashScreen::installIosSplashScreen()` builds `nativephp/ios/NativePHP/Assets.xcassets/LaunchImage.imageset/Contents.json` by checking `public/` for six hard-coded splash variants. When **all six exist** (the supported case), the generated Contents.json mixes `"Any"` scale entries (the unscaled `splash.png` / `splash-dark.png`) with specific-scale entries (`@2x`, `@3x`) — and Xcode's asset catalog compiler rejects that mix at archive time:

```
error: Image set has a child with bitmap content and the "Any" scale.
The image set also has children with specific scales and content.
Please remove the content with specific scales or change the content
of the "Any" scale to vector content.
** ARCHIVE FAILED **
```

The bug hides because **simulator builds skip `GenerateAssetSymbols`**, so the failure only surfaces on a release archive (Bifrost or `xcodebuild archive`).

## Affected

- Package: `nativephp/mobile@3.3.5` (latest at time of writing)
- File: `src/Traits/InstallsSplashScreen.php`
- Trigger: a Laravel app that provides all six splash files in `public/` (`splash.png`, `splash-dark.png`, `splash@2x.png`, `splash-dark@2x.png`, `splash@3x.png`, `splash-dark@3x.png`)

## Reproduction

1. Create a new NativePHP Mobile app.
2. Add all six splash variants to `public/`. Apple-recommended resolutions, doesn't matter — the bug is structural.
3. Run `php artisan native:install ios` (or let Bifrost run it).
4. Inspect `nativephp/ios/NativePHP/Assets.xcassets/LaunchImage.imageset/Contents.json` — it contains all six entries; the first two have no `scale` key.
5. Run `xcodebuild archive` (e.g., via `native:build ios` for an archive build, or any release configuration that runs the asset compiler). Bifrost reproduces this on every iOS App Store build with all six files present.
6. Archive fails with the error above.

## Root cause

`$splashVariants` in [`InstallsSplashScreen.php`](https://github.com/NativePHP/mobile-air/blob/main/src/Traits/InstallsSplashScreen.php):

```php
$splashVariants = [
    'splash.png'         => ['filename' => 'splash.png',         'idiom' => 'universal'],
    'splash-dark.png'    => ['filename' => 'splash-dark.png',    'idiom' => 'universal', 'appearances' => [['appearance' => 'luminosity', 'value' => 'dark']]],
    'splash@2x.png'      => ['filename' => 'splash@2x.png',      'idiom' => 'universal', 'scale' => '2x'],
    'splash-dark@2x.png' => ['filename' => 'splash-dark@2x.png', 'idiom' => 'universal', 'scale' => '2x', 'appearances' => [...]],
    'splash@3x.png'      => ['filename' => 'splash@3x.png',      'idiom' => 'universal', 'scale' => '3x'],
    'splash-dark@3x.png' => ['filename' => 'splash-dark@3x.png', 'idiom' => 'universal', 'scale' => '3x', 'appearances' => [...]],
];
```

The first two entries have no `scale` key (so Xcode treats them as `"Any"` scale). The trait appends an entry to `$images` for every variant whose file is found, then writes them all to Contents.json without checking that the resulting set is valid.

## Suggested fix (either of these would work)

**Option A — skip the unscaled variants when scaled variants are also present:**

```php
$hasScaledVariants = collect(['splash@2x.png', 'splash-dark@2x.png', 'splash@3x.png', 'splash-dark@3x.png'])
    ->contains(fn ($f) => File::exists(public_path($f)));

if ($hasScaledVariants) {
    unset($splashVariants['splash.png'], $splashVariants['splash-dark.png']);
}
```

This preserves existing behavior for projects with only unscaled splashes (single entry, `"Any"` scale, valid) while producing a valid Contents.json for projects with the full set.

**Option B — give the unscaled entries an explicit `'scale' => '1x'`:**

```php
'splash.png'      => ['filename' => 'splash.png',      'idiom' => 'universal', 'scale' => '1x'],
'splash-dark.png' => ['filename' => 'splash-dark.png', 'idiom' => 'universal', 'scale' => '1x', 'appearances' => [['appearance' => 'luminosity', 'value' => 'dark']]],
```

Simpler diff, but changes behavior for projects that previously relied on `splash.png` being the `"Any"` fallback for all densities (e.g., a project shipping a single high-res splash and letting iOS downscale).

Option A is safer in that respect; happy to PR whichever you prefer.

## Workaround for users hitting this today

Delete `public/splash.png` and `public/splash-dark.png`. The trait then generates a Contents.json with only `@2x` + `@3x` entries — no mix, Xcode accepts it. iOS handles a missing 1x by downsampling from 2x; no currently-supported iOS device uses 1x scale.

## Why this often goes unnoticed

Simulator builds skip `GenerateAssetSymbols`, so the broken Contents.json passes locally. Only release/archive builds (App Store submission, Bifrost) trigger the asset compiler and surface the failure. A project can ship simulator builds for months without seeing it.

barrymortenson:
Fixes #141.

## What

`InstallsSplashScreen::installIosSplashScreen()` builds `LaunchImage.imageset/Contents.json` from whichever of six hard-coded splash variants exist in `public/`. The first two (`splash.png`, `splash-dark.png`) have no `scale` key, so Xcode treats them as `"Any"` scale. When `@2x` and/or `@3x` variants also exist, the generated Contents.json has a mix Xcode's asset compiler rejects:

```
error: Image set has a child with bitmap content and the "Any" scale.
The image set also has children with specific scales and content.
Please remove the content with specific scales or change the content
of the "Any" scale to vector content.
** ARCHIVE FAILED **
```

Simulator builds skip `GenerateAssetSymbols` so the bug stays invisible until a release archive runs the asset compiler. That made it survive in our project until Bifrost's first App Store build hit it.

## How

If any scaled variant exists, drop the unscaled ones from the candidate set before the existence loop:

```php
$hasScaledVariants = false;
foreach (['splash@2x.png', 'splash-dark@2x.png', 'splash@3x.png', 'splash-dark@3x.png'] as $scaledFilename) {
    if (File::exists(public_path($scaledFilename))) {
        $hasScaledVariants = true;
        break;
    }
}
if ($hasScaledVariants) {
    unset($splashVariants['splash.png'], $splashVariants['splash-dark.png']);
}
```

## Behavior matrix

| `public/` contains | Before | After |
|---|---|---|
| only `splash.png` (± `-dark`) | one "Any"-scale entry (valid) | unchanged — one "Any"-scale entry (valid) |
| only `@2x` and `@3x` variants | scaled-only entries (valid) | unchanged — scaled-only entries (valid) |
| all six variants | mixed Any + scaled (invalid → archive fails) | scaled-only entries (valid) |

The 1x splash files are effectively ignored when scaled variants are present. iOS handles a missing 1x by downsampling from 2x; no currently-supported iOS device uses 1x scale anyway (the original iPhone is the only 1x device and is unsupported by modern iOS).

## Alternative considered

I considered just adding `'scale' => '1x'` to the two unscaled entries in `$splashVariants`. Simpler diff, but it changes the meaning for projects that ship a single `splash.png` and rely on it being the "Any" fallback across densities — `scale: 1x` would constrain it to 1x devices only. The conditional-skip approach in this PR avoids that behavior change.

## Test plan

- [x] Manually verified with all six variants: generated Contents.json now contains only `@2x` and `@3x` entries; archive succeeds.
- [x] Manually verified with only `splash.png` (no scaled variants): generated Contents.json contains the single unscaled entry; archive succeeds.
- [ ] No existing iOS splash unit tests in this repo (`tests/Unit/Traits/InstallsAndroidSplashScreenTest.php` covers Android only). Happy to add an `InstallsIosSplashScreenTest` in a follow-up if you'd like — the Android test gives a clear pattern to follow.

### DIFF
--- a/src/Traits/InstallsSplashScreen.php
+++ b/src/Traits/InstallsSplashScreen.php
@@ -18,6 +18,24 @@ trait InstallsSplashScreen
             'splash-dark@3x.png' => ['filename' => 'splash-dark@3x.png', 'idiom' => 'universal', 'scale' => '3x', 'appearances' => [['appearance' => 'luminosity', 'value' => 'dark']]],
         ];
 
+        // Xcode's asset compiler rejects Image Sets that mix entries with no
+        // `scale` key ("Any" scale) and entries with specific scales (1x/2x/3x)
+        // when the "Any" entries reference bitmap content. If the project ships
+        // any scaled variants, drop the unscaled ones so the generated
+        // Contents.json doesn't trigger:
+        //   error: Image set has a child with bitmap content and the "Any"
+        //   scale. The image set also has children with specific scales...
+        $hasScaledVariants = false;
+        foreach (['splash@2x.png', 'splash-dark@2x.png', 'splash@3x.png', 'splash-dark@3x.png'] as $scaledFilename) {
+            if (File::exists(public_path($scaledFilename))) {
+                $hasScaledVariants = true;
+                break;
+            }
+        }
+        if ($hasScaledVariants) {
+            unset($splashVariants['splash.png'], $splashVariants['splash-dark.png']);
+        }
+
         $foundVariants = [];
         $images = [];
 


simonhamp (approved):
Thanks for reporting and supplying a fix

PR #143, opened 2026-05-23
Fix repeated query parameters in Android WebView requests

buseduran:
This PR fixes Android request forwarding so repeated query parameters are preserved when NativePHP forwards WebView requests to Laravel.

Previously, query parameters were converted through a `Map`-based structure, which collapsed duplicate keys such as `symptoms[]` into a single value. This caused multi-select form fields to lose data on Android even though the backend and form submission logic were working correctly.

### Changes
- Preserve the raw encoded query string in `PHPRequest`
- Use `request.url.encodedQuery` in `PHPWebViewClient` for PHP requests and asset fallbacks
- Preserve query strings when handling redirects with absolute URLs
- Keep backward compatibility by retaining `getParameters` as a fallback

### Validation
- Verified on Android emulator that multi-select checkbox values now arrive correctly
- Confirmed Laravel/backend persistence still stores multiple symptom values
- Local regression tests passed

### DIFF
--- a/resources/androidstudio/app/src/main/java/com/nativephp/mobile/network/PHPRequest.kt
+++ b/resources/androidstudio/app/src/main/java/com/nativephp/mobile/network/PHPRequest.kt
@@ -6,16 +6,20 @@ data class PHPRequest(
     val body: String = "",
     val headers: Map<String, String> = emptyMap(),
     val getParameters: Map<String, String> = emptyMap(),
+    val queryString: String = "",
     val postParameters: Map<String, String> = emptyMap(),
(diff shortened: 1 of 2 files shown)


simonhamp:
Thanks @buseduran! Do you know if this is an issue on iOS too?

buseduran:
Thanks for the review! I checked the iOS path as well, and the original Android Map-collapse issue does not seem to exist there. iOS preserves the raw percent-encoded query string for the initial request flow, so repeated parameters are handled correctly.

simonhamp:
Great! Thanks for confirming

simonhamp (approved):

PR #146, opened 2026-05-25
padding-bottom prop for iOS 26+ Tab bar with Liquid Glass

ArthurYdalgo:
## Summary

> [!NOTE]  
> I added a disclaimer by the end of this PR description to explain that, upon further testing, this new feature (although "usable") probably doesn't need to exist as it was just a Bifrost Jump visual behavior

I was testing NativePHP on an iOS device with Liquid Glass through Jump and I noticed the [Bottom Navigation bar](https://nativephp.com/docs/mobile/3/edge-components/bottom-nav) was "too high" (to the point of being in the way of some visual elements)

Screenshot 2026-05-25 at 17 07 20

## Suggested Change

A new 'padding-bottom' property that can be used in the Bottom Navigation EDGE component to move the tab bar vertically if needed, for iOS 26+ Tab bar with Liquid Glass only. The default 20 is still used in case it is not set.

### AI Usage

The original changes were made with Gemini 3.1 Pro and additional changes were made by a human (me).

### Wether or not it is relevant

After I was done with this PR, I tried running the `php artisan native:watch ios` and tested on the iOS emulator (previously I had only tried using Bifrost Jump), and realized the Bottom Nav only shows up at the height it does on the previous screenshot when running on a real device through Jump (or at least that's my conclusion so far). 

So this entire PR became pointless but I had already went through the trouble trying to do it and testing it out so I thought I might as well submit it anyway in case anybody ever need this one day and someone chooses to reopen this PR.

The screenshot below is from when I tried running the same project though `native:watch`

simulator_screenshot_4BBD61AE-F8C7-484B-BFA6-04A5E05F3086

### DIFF
--- a/resources/xcode/NativePHP/NativeUI/NativeBottomNav.swift
+++ b/resources/xcode/NativePHP/NativeUI/NativeBottomNav.swift
@@ -271,6 +271,8 @@ struct NativeBottomNavigation: View {
     var body: some View {
         if let bottomNavData = uiState.bottomNavData,
            let items = bottomNavData.children, !items.isEmpty {
+           
+            let customBottomPadding = CGFloat(bottomNavData.paddingBottom ?? 20)
 
             if #available(iOS 26.0, *) {
                 // iOS 26+: Tab bar with Liquid Glass effect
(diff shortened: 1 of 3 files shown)


ArthurYdalgo:
As mentioned in the PR description, I don't really think this is relevant so I'll just close it myself and leave it here in case it ever becomes needed again.

PR #156, opened 2026-06-06
fix(android): avoid resizing WebView for keyboard insets

buseduran:
This PR fixes a keyboard animation glitch on Android by avoiding IME inset padding on the WebView container.

Previously, applying `WindowInsets.ime` padding to the `AndroidView` resized the WebView while the keyboard animated in. That forced web layouts using viewport units or fixed positioning to reflow during the animation, causing visible jumps.

The activity still uses `adjustResize`; this only removes the extra Compose IME padding from the WebView container. The WebView continues to respect the Scaffold top/bottom padding, while Chromium handles the visual viewport and focused-field scrolling during keyboard input.

### Changes
- Keep the WebView full-size while still respecting top and bottom navigation padding
- Let Chromium adjust the visual viewport and scroll the focused field into view
- Document why IME padding is intentionally not applied to the WebView container

### Validation
- Verified on Android emulator that focusing an input opens the keyboard without the previous WebView jump

### DIFF
--- a/resources/androidstudio/app/src/main/java/com/nativephp/mobile/ui/MainActivity.kt
+++ b/resources/androidstudio/app/src/main/java/com/nativephp/mobile/ui/MainActivity.kt
@@ -40,7 +40,6 @@ import androidx.compose.foundation.Image
 import androidx.compose.foundation.background
 import androidx.compose.foundation.isSystemInDarkTheme
 import androidx.compose.foundation.layout.*
-import androidx.compose.foundation.layout.ime
 import androidx.compose.material3.ColorScheme
 import androidx.compose.material3.MaterialTheme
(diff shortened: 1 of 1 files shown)


simonhamp:
@buseduran could you fix the conflicts and check that this still works as intended against v4?

buseduran:
@simonhamp rebased onto v4, conflict solved and the change itself is the same two lines.

Re-tested against v4 and the regression is still there without the patch: focusing an input resizes the layout viewport repeatedly during the keyboard animation, so 100vh / position: fixed layouts reflow and jump. With the patch innerHeight stays put and only the visual viewport changes, with Chromium still scrolling the focused field into view.

simonhamp:
Thanks!

PR #158, opened 2026-06-09
Fix: Remove stray tag

martin-ro:
Removes `</laravel-boost-guidelines>` from Boost guideline which is causing duplicate entries on each `boost:update`.


simonhamp:
Thanks

PR #159, opened 2026-06-09
Better splash screens: iOS SVG + continuous native Android splash

unlocdavid:
Full disclosure: This PR is made with Claude Opus

I had some issues with splash screens. On Android the splash was fuzzy in some cases, maybe because of my UI scaling. And the native Android splash and NativePHP splash generated visual noise on boot.

I investigated if SVG would be an option, but it seems to only be reliable for iOS. So this PR adds support for `splash.svg` and `splash-dark.svg` only on iOS. If an SVG is found, it takes precedence over PNGs.

On Android 12+ there is a native splash screen of the app icon in a circle on a colored background. After the first loading of the app it jumped to the NativePHP configured PNG splash which waits for PHP etc. boots. This PR adds the option to hold the native splash and don't show the PNG. This solves two problems:

- No fuzzy splash screens.
- No jarring jump between splash screens. Especially in a release build, non-first boots can be very fast (yay!) but jumping between the two screens generates a lot of visual noise for the user. Using `native` in the config eliminates the jump.  

Below is the description generated by Claude:

---

## Summary

Two splash-screen improvements:

- **iOS** accepts a native vector `splash.svg` (no rasterizer).
- **Android** holds the system splash until the WebView is ready, eliminating the
  unbranded-system-splash → custom-overlay jump, with a new config to choose the style.

Android splash input remains PNG-only (an SVG→VectorDrawable path was explored and dropped as too unreliable).

## iOS

- `public/splash.svg` (+ optional `splash-dark.svg`) is written into `LaunchImage.imageset`
  as a vector asset (`preserves-vector-representation`). Invalid/missing SVG falls back to
  the existing PNG pipeline.

## Android

- Add `androidx.core:core-splashscreen`; `installSplashScreen()` +
  `setKeepOnScreenCondition` hold the system splash through boot in `native` style.
- Generate `Theme.AndroidPHP.Splash` and point the app theme at it.
- New `android.splash` config: `style` (`image` default | `native`), `background`,
  `background_night` (env: `NATIVEPHP_ANDROID_SPLASH_*`).
- The system splash shows the **app launcher icon** in both styles. `image` keeps the
  full-bleed Compose overlay; `native` relies on the held system splash on API 31+, with
  the overlay as the fallback below 31.
- Splash style/icon/background are applied at **build time**, so config changes take
  effect on rebuild (not just install).
- Post-splash `windowBackground` tracks the splash color to avoid a white flash on handoff.
- A `@drawable/splash` drawable is always guaranteed (shipped placeholder + app-icon/PNG
  fallback) so resource linking never fails.

## Config

  ```php
  'splash' => [
      'style' => env('NATIVEPHP_ANDROID_SPLASH_STYLE', 'image'),
      'background' => env('NATIVEPHP_ANDROID_SPLASH_BACKGROUND', '#FFFFFF'),
      'background_night' => env('NATIVEPHP_ANDROID_SPLASH_BACKGROUND_NIGHT', '#000000'),
  ],
```
Old configs without this block fall back to image mode with white/black defaults.


simonhamp:
Hey there. Thanks for your work on this PR. We're going to leave this functionality to [Plugins](https://nativephp.com/plugins) rather than adding more complexity to the core.

If there's some aspect of core that would make it easier for you to build a plugin for this feature, we would welcome it. Please make this clear in any future PR description.

PR #161, opened 2026-06-09
Fix Jump hot reload ignoring nativephp.hot_reload.watch_paths (Issue #160)

### LINKED ISSUE #160: Jump hot reload ignores nativephp.hot_reload.watch_paths

Use php artisan native:jump for on-device development with hot reload, and have it respect the same hot_reload.watch_paths config as php artisan native:watch.

I added custom directories (e.g. modules/) to config/nativephp.php → hot_reload.watch_paths so changes there would reload the app on my phone while developing with Jump.

### What happened?

Changes under custom watch paths did not trigger a reload when using Jump, even though:

The same paths do reload with native:watch (Android/iOS device builds)
Changes under the default paths (app/, resources/, routes/, config/) still reload via Jump
Jump’s bridge file watcher uses a hardcoded list in resources/jump/websocket-server.php and never reads config('nativephp.hot_reload.watch_paths').

### How to reproduce the bug

In a Laravel app with NativePHP Mobile, publish/config and extend watch paths:
// config/nativephp.php
'hot_reload' => [
    'watch_paths' => [
        'app',
        'resources',
        'routes',
        'config',
        'modules', // custom directory
    ],
    // ...
],
Create a test file, e.g. modules/Foo/Example.php or a Blade view under modules/.

Run php artisan native:jump, connect a device, and keep the app open.

Edit and save a .php or .blade.php file under modules/.

Expected: device reloads (same as native:watch).

Actual: no reload.

Control: edit routes/web.php or resources/views/... — reload does happen via Jump.

### Debug Output

From storage/logs/jump-bridge.log after starting Jump — the bridge never logs custom paths; only the hardcoded set is used internally:

=== 2026-06-09 12:00:00 bridge server starting (ws=3001 tcp=3002 vite_proxy=3003) ===
[12:00:01] [Jump] TCP bridge listening on 127.0.0.1:3002
After editing modules/SomeModule/Example.php — no lines like:

Changed: modules/SomeModule/Example.php
Sent reload to 1 device(s)
After editing routes/web.php — reload does appear:

[12:05:10] [Jump] Changed: routes/web.php
[12:05:10] [Jump] Sent reload to 1 device(s)
Root cause (for maintainers): resources/jump/websocket-server.php hardcodes:

$watchPaths = ['app', 'resources', 'routes', 'config'];
native:watch reads config via WatchesAndroid / WatchesIos:

$paths = config('nativephp.hot_reload.watch_paths', $this->androidWatchPaths);
Jump’s bridge process is spawned from JumpCommand and does not receive Laravel config unless it is passed in (e.g. env var from JumpCommand).

Windows

Jump app (Android)

### Notes

_No response_

mylesduncanking:
Fixes #160

## Summary
Jump's bridge file watcher used a hardcoded path list (`app`, `resources`, `routes`, `config`) and ignored `config/nativephp.php` → `hot_reload.watch_paths`.
`native:watch` already reads this config (see `WatchesAndroid` / `WatchesIos`), so custom directories like `modules/` reload on device builds but not via Jump.

## Changes
- Pass `nativephp.hot_reload.watch_paths` from `JumpCommand` to the bridge server via `JUMP_WATCH_PATHS`
- Read that env in `websocket-server.php` with the existing paths as fallback
- Log resolved watch paths on bridge startup for easier debugging

### DIFF
--- a/resources/jump/websocket-server.php
+++ b/resources/jump/websocket-server.php
@@ -189,7 +189,8 @@ $wsWorker->onWorkerStart = function () use (&$deviceConnections, &$pendingCalls,
     // page props, scroll position). Only watch them when Vite isn't running.
     $lastModTimes = [];
     $lastReloadTime = 0;
-    $watchPaths = ['app', 'resources', 'routes', 'config'];
+    $watchPaths = jumpResolveWatchPaths();
+    jumpLog('File watcher paths: '.implode(', ', $watchPaths));
     $serverExtensions = ['php', 'blade.php'];
     $clientExtensions = ['js', 'jsx', 'ts', 'tsx', 'vue', 'css', 'scss', 'sass', 'less'];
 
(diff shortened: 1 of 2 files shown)


simonhamp:
@mylesduncanking could you fix the conflicts here and test against Jump v3, which was just released?

PR #163, opened 2026-06-13
Fix Windows path separator bug in registerNativeComponents()

rdanklof:
## Summary
   
Some developers are required to work on Windows due to company policy. When testing via the Jump app on Windows, native components (e.g. \`<native:bottom-nav>\`) failed to register because \`SplFileInfo::getPathname()\` returns backslashes, while the base path was stripped using a hardcoded \`/\` separator.

Fixed by normalizing both paths with \`str_replace('\\\\', '/')\` and using \`substr\` + \`strlen\` to extract the relative path.
   
## Test plan
   
- [ ] On Windows: open the Jump app and verify native components render correctly
- [ ] Verify no regression on macOS/Linux


tabatii:
when will this PR be merged ?
this bug prevents windows users from using edge components entirely

simonhamp:
Has anyone tested this?

rdanklof:
After the v4 release, I discovered this fix was required in several additional locations across the codebase. I've refactored it into a dedicated Support class to make it reusable and maintainable.

rdanklof:
> when will this PR be merged ? this bug prevents windows users from using edge components entirely

@tabatii Any chance you could test this on a Windows machine?

PR #164, opened 2026-06-14
Fix release bundle cache handling

WesWeCan:
## Summary
- Exclude Laravel `bootstrap/cache` from iOS and Android release bundle copies so stale dev package manifests are not reused after `composer install --no-dev`.
- Recreate the empty `bootstrap/cache` directory before Composer runs so Laravel package discovery can regenerate clean metadata.
- Make iOS fail fast when Composer install fails, matching Android's existing behavior.

## Why
Release builds install Composer dependencies with `--no-dev`, but the bundle copy could include cached `bootstrap/cache/packages.php` and `services.php` generated from the developer dependency set. If those cached files reference dev-only providers, Composer's post-autoload scripts can fail or a broken iOS bundle can be shipped.

## Test plan
- `./vendor/bin/pest ./tests/Feature/ReleaseBuildBundleTest.php`
- `./vendor/bin/pint --dirty --format agent`

## Notes
- The fix does not copy compiled cache files into the release bundle; it only preserves the required empty `bootstrap/cache` directory.
- Android already checked Composer failures, so the Android change is limited to cache exclusion and directory recreation.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/Commands/BuildIosAppCommand.php
+++ b/src/Commands/BuildIosAppCommand.php
@@ -92,7 +92,7 @@ class BuildIosAppCommand extends Command
         file_put_contents($this->appPath.'.env', PHP_EOL.'ASSET_URL="/_assets"'.PHP_EOL, FILE_APPEND);
 
         $this->components->task('Installing Composer dependencies', function () {
-            Process::path($this->appPath)
+            $result = Process::path($this->appPath)
                 ->forever()
                 ->run([
                     'composer',
@@ -105,6 +105,19 @@ class BuildIosAppCommand extends Command
                         $this->output->write($output);
                     }
                 });
+
+            if (! $result->successful()) {
+                $errorOutput = $result->errorOutput();
+                if ($errorOutput) {
+                    file_put_contents($this->logPath, $errorOutput, FILE_APPEND);
+                }
+
+                error('Composer install failed. Check your dependencies and try again.');
+                file_put_contents($this->logPath, 'ERROR: composer install failed with exit code '.$result->exitCode().PHP_EOL, FILE_APPEND);
+                exit(1);
+            }
+
+            return true;
         });
 
         $this->components->task('Removing unnecessary files', fn () => $this->removeUnnecessaryFiles());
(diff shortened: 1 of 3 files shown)


shanerbaner82:
Thanks for this — the fix itself is solid and we want it. The stale `bootstrap/cache` failure under `composer install --no-dev` is real, and the missing composer exit-code check on iOS is a genuine ship-a-broken-bundle bug. The `src/` changes look correct as-is.

A few things needed before we can merge:

**1. Rebase onto `main`**

The PR currently targets `element`, and the branch was cut from it — retargeting alone would drag ~227 unrelated commits into the diff. Please rebase your three file changes onto `main` (or open a fresh branch from `main` and cherry-pick). I've confirmed both touched files are identical on `main` in the regions you're patching, so it'll apply cleanly.

**2. Remove `test_ios_composer_install_result_is_checked`**

This test reads the command's source file and asserts it contains literal strings like `$result = Process::path($this->appPath)`. That tests the text of the code rather than its behavior, and it'll break on any refactor or formatting change. Either drop it, or replace it with a `Process::fake()`-based test that asserts the failure path actually triggers.

**3. Slim down the Android test (optional but preferred)**

`test_android_release_bundle_excludes_cached_bootstrap_files_and_recreates_cache_directory` runs a real `composer install`, which requires composer on the CI PATH and adds runtime/flake risk. Also worth noting: `ReleaseBuildTester` reimplements `platformOptimizedCopy` in PHP, so the test never exercises the real rsync/robocopy exclusion — the part most likely to misbehave. If you can assert on the temp-dir/zip contents without invoking real composer, that'd make this much more CI-friendly. Not a hard blocker.

**One heads-up, no action needed:** on Windows, the copy-time exclusion goes through robocopy as `/XD "{source}\bootstrap/cache"` with a mixed-slash path — and Windows is the one platform where the zip step doesn't also exclude `bootstrap/cache`, so this exclusion is load-bearing there. If you have access to a Windows machine to verify robocopy actually skips it, that would be great; otherwise we'll check it on our side.

Happy to merge once the rebase and the test removal are in. 🙏

WesWeCan:
Thanks for the review. I addressed the feedback in #194: the replacement is a single commit on current main, the source-inspection iOS test is removed, and the Android test now fakes Composer while asserting the exclusion and ZIP contents. I do not have Windows access, so the robocopy /XD behavior remains for your side to verify as offered. Closing this element-based PR in favor of #194.

PR #170, opened 2026-06-24
fix / side nav header status bar inset

omaralalwi:
Side-nav drawer header is clipped under the status bar — the drawer renders edge-to-edge but its content doesn't consume the top safe-area inset


simonhamp:
Thanks. Can you provide screenshots of before and after on iOS and Android?

simonhamp:
Closing due to inactivity

PR #174, opened 2026-07-12
Add WidgetKit extension target support for plugins

Matildevoldsen:
## Summary

- add a validated `ios.extension_targets` manifest API, initially scoped to WidgetKit app extensions
- isolate extension Swift sources from the host target and generate deterministic Xcode targets, dependencies, products, entitlements, Info.plists, and embed phases
- share App Group entitlements with the host and widget extension while preserving existing host entitlement values
- support automatic local signing and per-bundle manual provisioning profiles for archive/export builds
- clean only compiler-owned sources and Xcode blocks, including safe recovery after interrupted compilation

## Why

WidgetKit code runs in a separate process and must be built as its own `com.apple.product-type.app-extension` target. Plugins can currently copy Swift into the host app, but they cannot declare or embed that additional target. This implements the core portion requested in [NativePHP discussion #669](https://github.com/orgs/NativePHP/discussions/669).

## Manifest API

```json
{
  "ios": {
    "min_version": "17.0",
    "entitlements": {
      "com.apple.security.application-groups": [
        "group.${APP_ID}.widgets"
      ]
    },
    "extension_targets": [
      {
        "name": "NativePHPWidgetsExtension",
        "type": "widget-extension",
        "bundle_id_suffix": "widgets",
        "sources_dir": "extension",
        "info_plist": {
          "CFBundleDisplayName": "NativePHP Widgets"
        }
      }
    ]
  }
}
```

`deployment_target` is optional and falls back to `ios.min_version`, then iOS 17. The compiler injects the WidgetKit extension point and owns bundle/version/package plist keys so manifests cannot create signing mismatches.

For manual distribution signing, `IOS_EXTENSION_PROVISIONING_PROFILES` accepts a JSON object keyed by each full extension bundle ID. Values may be readable `.mobileprovision` paths or strict base64. Profiles are CMS-decoded, checked against the exact `application-identifier`, installed by UUID, applied to the generated target, and added to `ExportOptions.plist`.

## Safety and packaging

- generated target directories are recorded in an ownership manifest; unrelated `ios/Extensions` content is preserved
- stale extension entry points are removed from the host compiler output
- project edits use deterministic managed blocks and are byte-idempotent, including after `pod install`
- archives containing `.appex` bundles use `xcodebuild -exportArchive` so Xcode preserves nested signatures and entitlements
- cleanup restores generated extension targets to automatic signing without touching unrelated custom targets

## Verification

- PHP 8.3 and PHP 8.4: **326 passed**, 3 skipped, 4 pre-existing risky tests, **978 assertions**
- Laravel Pint: passed
- targeted PHPStan over every changed production path: no errors
- `plutil` validation: generated host/extension entitlements and extension Info.plist passed
- `pod install` after target injection: target preserved and repeated compilation remained byte-identical
- Xcode 26.6 / iOS 26.5 SDK:
  - full `NativePHP-simulator` workspace build passed on iPhone 17 Pro Max simulator
  - extension embedded at `PlugIns/NativePHPWidgetsExtension.appex`
  - PluginKit registered `com.matildeene.laraloom.widgets`
  - widget appeared in the gallery and rendered on the simulator Home Screen
  - unsigned Release `iphoneos` extension build produced an arm64 Mach-O

The manual profile installation/export boundary is integration-tested with decoded profile fixtures. A live App Store export was not attempted because that requires matching production host and extension profiles.

## Scope

This draft intentionally covers iOS WidgetKit target support only. The Android receiver metadata and resource-copy primitives described by the original widget proposal are already present in current core. Once the manifest shape is agreed, I can follow with the separate documentation PR requested by the contribution guide.


simonhamp:
Hey @Matildevoldsen, there's now a conflict as some changes from #173 came over.

Would love to get this one in place though - would be awesome!

shanerbaner82:
Thanks @Matildevoldsen — this is a seriously well-put-together PR. The Apple mechanics are right (`dstSubfolderSpec = 13` embed, `SKIP_INSTALL`, `APPLICATION_EXTENSION_API_ONLY`, falling back to `xcodebuild -exportArchive` when an `.appex` is present), the deterministic IDs + managed blocks make the project edits idempotent and reversible, and testing against the real template is exactly what I'd want for something this invasive. The manifest shape also looks right — scoping to `widget-extension`, deriving bundle IDs from a suffix, and locking the compiler-owned plist keys are the right calls, so consider the shape agreed.

A few things I'd like addressed before this merges. The first is a security concern:

**1. `${ENV_VAR}` expansion needs an allowlist (security)**

`ManifestValueResolver` currently expands *any* `${VAR}` found in a plugin manifest into Info.plist/entitlements via `env()`. Plugin manifests are third-party content, so this is a channel for a malicious or compromised plugin to embed arbitrary build-machine environment values (CI signing secrets, tokens) into a shipped artifact — or simply read them. Please restrict expansion to `APP_ID` plus explicitly blessed variables, consistent with the posture `PluginSecretsValidator` takes.

**2. Recovery after Xcode re-saves the project**

If a user opens `nativephp/ios` in Xcode and Xcode rewrites `project.pbxproj` (which it does on any UI-driven change), the `/* NATIVEPHP_EXTENSIONS_* */` markers are dropped while the inserted objects remain. On the next compile, `stripManagedBlocks` finds nothing to strip and `assertTargetIsAvailable` throws "already contains a target named …", with no way out short of hand-editing the project. Since your object IDs are deterministic, recovery seems tractable: also strip any objects whose 24-char IDs match the hash scheme, not just the marker blocks. Users open this project in Xcode all the time, so I'd treat this as a blocker too.

**3. Make the no-extension path a no-op**

Removing the `count === 0` early return means every iOS build now reads and rewrites the pbxproj even for the majority of apps with no extensions. A cheap `str_contains($contents, 'NATIVEPHP_EXTENSIONS_BEGIN')` guard before writing would keep the hot path read-only for everyone not using this feature.

**Worth discussing: the pbxproj editing strategy**

The regex editor is carefully done, but it's ~550 lines of surgery keyed to the template's exact formatting, now on every user's build path. Before we commit to owning that long-term, I'd like to weigh two alternatives: (a) the `xcodeproj` Ruby gem — CocoaPods already puts Ruby on every machine that builds, and it's the same library CocoaPods itself uses to mutate the project safely; (b) shipping a dormant extension target in the template so the compiler only fills in sources/settings rather than doing structural surgery. If we keep the regex approach, the current safeguards make it defensible — I just want that to be a deliberate choice.

**Minor**

- A comment on the `putenv(NATIVEPHP_IOS_EXTENSION_PROVISIONING_PROFILES)` channel would help future readers — it works because packaging spawns `native:build` as a child process that inherits the environment, which is easy to break unknowingly.
- `PropertyList::decode` throws on `<date>`/`<data>`, and `mergeEntitlements` now decodes the host entitlements on every build that touches plugin entitlements — either support those types or fail with a message that names the file.
- The entitlements merge quietly changes behavior (existing host scalars now win over plugin values; lists union). I think that's the better semantic, but it deserves a changelog note.
- The comment-deletion churn in `Plugin.php` is unrelated — please drop it to keep the diff focused.

On the Android side, I've confirmed current core already covers widget needs (receivers with nested `meta-data android:resource` and intent filters, plus `android.assets` copying into `res/xml`/`res/layout`), so agreed: no Android work needed in this PR.

Happy to see the docs PR once these are addressed. Thanks again — this is one of the most thorough contributions we've had.

PR #208, opened 2026-07-22
Remove duplicate runtime blocks from nativephp config

martin-ro:
The 'runtime' key was defined three times (last one silently won). Collapsed into one block and moved the orphaned cleanup_env_keys doc comment back above its array. No effective config changes.


simonhamp (approved):
Thanks. Some weird merge stuff going on somewhere

PR #210, opened 2026-07-22
fix(ios): stop mangling certificate passwords containing special characters

shanerbaner82:
## The bug

A user hit `Failed to import certificate` / `Failed to setup signing credentials` on `native:package ios`. Their `.p12` password contained a `#`. Re-exporting the certificate with an alphanumeric password fixed it.

## Root cause

`native:credentials` wrote the password to `.env` **unquoted**. phpdotenv treats an unquoted `#` as the start of a comment, so the value is silently truncated before it ever reaches `security import`. Verified against this repo's own `vendor/vlucas/phpdotenv`:

| raw `.env` line | parsed value |
|---|---|
| `PW=abc#123` | `abc` — silently truncated |
| `PW=abc 123` | throws `Failed to parse dotenv file` |

The space case is worse than a build failure: it throws during bootstrap, so no artisan command runs at all.

The tool was mangling its own output, and the failure was then reported as a bare `Failed to import certificate` — `security`'s actual stderr (`The user name or passphrase you entered is not correct.`) was thrown away, which is why this took several rounds to diagnose.

## Changes

1. **`CreatesIosCredentials::updateIosEnvVars()`** — values are double-quoted with `\`, `"` and `$` escaped. 57 adversarial passwords (every ASCII punctuation character individually, `${HOME}`, backslashes, quotes, leading/trailing spaces, unicode, 200-char) round-trip byte-for-byte through real phpdotenv. `IOS_DISTRIBUTION_CERTIFICATE_PATH` still parses to `credentials/distribution.p12` with no quotes attached.
2. **`ManagesIosSigning::installCertificate()`** — prints the real `security import` stderr on failure, then names the offending `.env` variable if its value is unquoted and contains `#`. **Only `strlen($password)` is ever printed, never the secret.**
3. **Entry-time warnings** — `native:credentials` warns if the chosen password needs quoting, and warns when it leaves an existing password variable untouched (it never overwrites, which is how a bad value survives a re-run).

## Verification

- Reproduced end-to-end with a real generated `.p12` and password `p@ss#w0rd`: the truncated password produces exactly the reported failure; the full password verifies against the `.p12`.
- Diagnostic scanner covered for duplicate declarations (dotenv resolves the **last** assignment, so that is the one inspected), commented-out lines, and prefix collisions — 8/8.
- Full suite green: **602 passed, 0 failures** (4 risky / 3 skipped are pre-existing).
- Pint clean.

## Blast radius

Checked deliberately, since this touches signing:

- **Bifrost is unaffected.** It has zero references to `IOS_DISTRIBUTION_CERTIFICATE_PATH`/`PASSWORD` — it sources the certificate and password from `credentials.json` and passes them as CLI flags, bypassing `.env` entirely. Its build-log scraper matches fixed literals, none of which collide with the new diagnostic text, and the success path is byte-identical.
- Both traits are only consumed by real `Illuminate\Console\Command` subclasses, so `$this->line()` / `$this->components` are always available. No trait method-name collisions.
- No other code in this repo reads these variables with a raw regex where a quoted value would now be read with quotes attached.

## Separate issue, not fixed here

`bifrost`'s `build.sh` builds `--certificate-password="$CERT_PASSWORD"` into a shell string run through `eval`. A password containing `$`, a backtick or `"` gets re-expanded there. That is a pre-existing bug on a path this PR does not touch, but it is the same class of problem and worth a follow-up.


src/Concerns/CreatesIosCredentials.php
```diff
@@ -276,6 +276,16 @@ private function handleCertificateConversion(string $credentialsDir, string $key
...
+        if (preg_match('/[#\s"\'\\\\$!`]/', $p12Password)) {
+            $this->newLine();
+            $this->warn('⚠️  That password contains characters that need quoting.');
+            $this->line('   It will be written to .env double quoted, which is safe, but if you');
+            $this->line('   ever pass it via --certificate-password, wrap it in single quotes so');
+            $this->line('   your shell does not alter it before PHP sees it.');
+            $this->line('   An alphanumeric password avoids the issue entirely.');
+            $this->newLine();
+        }
+
```
simonhamp, line 288: don't think we need this at all

```suggestion

```

src/Concerns/CreatesIosCredentials.php
```diff
@@ -441,6 +451,27 @@ private function updateIosEnvVars(string $p12File, string $p12Password): void
...
             }
+
+            if (in_array($passwordVar, $existingVars, true)) {
+                $this->newLine();
+                $this->warn("⚠️  {$passwordVar} was not overwritten. If your password contains special");
+                $this->line('   characters, make sure the existing value is wrapped in double quotes:');
+                $this->line("       {$passwordVar}=\"your#password\"");
+                $this->line('   An unquoted # is treated as the start of a comment and silently');
+                $this->line('   truncates the password, which fails the keychain import at build time.');
+            }
```
simonhamp, line 462: equally I think remove

```suggestion
```

simonhamp (approved):
My suggestions are just that. Accept or don't, this is fine to me

I will action the Bifrost build script note too

simonhamp:
@shanerbaner82 there are conflicts here now that need resolving locally

shanerbaner82:
Re-reviewed and refreshed this branch against current `main`:

- **Merged main in** (`c3cf17d`) — the branch predated the `src/Traits` → `src/Concerns` rename, and GitHub's rename detection gave up (PR showed CONFLICTING). Git's local merge carried both edits into the new paths cleanly; full suite back at main's baseline (859 passed, same risky/skipped), PHPStan baseline unchanged, Pint clean.
- **Independently re-verified the core claim**: 42 adversarial passwords (every ASCII punctuation char, `${HOME}`, `$HOME`, backslash runs, mixed quotes, leading/trailing spaces, unicode, 200-char, `p@ss#w0rd`) round-trip byte-for-byte through the vendored phpdotenv via `quoteEnvValue()`, and `IOS_DISTRIBUTION_CERTIFICATE_PATH` still parses without quotes attached.
- **Follow-up found**: `CreatesAndroidCredentials::updateAndroidEnvVars()` (src/Concerns/CreatesAndroidCredentials.php:430) writes `ANDROID_KEYSTORE_PASSWORD` / `ANDROID_KEY_PASSWORD` with the same unquoted `{$key}={$value}` pattern — the identical `#`-truncation bug on the Android side. Worth a sibling fix that shares `quoteEnvValue()`.
- Nit for a follow-up rather than this PR: the round-trip verification lives outside the suite — a small permanent test for `quoteEnvValue()` would keep the escaping honest if phpdotenv's parsing ever changes.

PR #215, opened 2026-07-23
fix(plugins): copy Android plugin sources into a dedicated wiped source root

simonhamp:
Follow-up to #188, closing the known gap called out there.

## Problem

#188 wipes the generated tree before each compile, but Android plugin sources with a `package` declaration were copied to package-derived paths under `app/src/main/java` — *outside* that tree, interleaved with core app code — so they were never pruned. Renaming or removing a file in a plugin (or uninstalling the plugin) left a stale copy behind that re-declared its classes and broke the Gradle build (`Redeclaration` / ambiguous-declaration errors).

A directory wipe there could never be made safe: `src/main/java` is durable, user-owned territory, and nothing prevents plugin packages from interleaving with hand-written code.

## Fix

Make Android work like iOS: **one compiler-owned directory, wiped on every compile.** All compiler output now goes into a dedicated Gradle source root at `app/src/nativephp/kotlin/`:

- Plugin copies land at their package-derived path *inside* the root (`src/nativephp/kotlin/com/acme/camera/Camera.kt`). Kotlin doesn't require path-mirrors-package, and keeping the layout relative to a registered source root keeps Android Studio navigation/inspections behaving exactly as before.
- The generated `PluginBridgeFunctionRegistration.kt` / `PluginRendererRegistration.kt` moved in there too (same Kotlin package), so `clean()` is a single directory delete and the invariant is simply *"the compiler owns `src/nativephp/`"*.
- The template's `build.gradle.kts` declares the root via `java.srcDir("src/nativephp/kotlin")`; the template's placeholder registrations moved to match.

Renames, deletions, and uninstalls all resolve identically now — every compile starts from an empty root. No manifest/bookkeeping needed.

## Upgrades self-heal

Two migration mechanisms run on every compile (no-ops once clean), so projects upgraded via `composer update` alone — without re-running `native:install` — keep building:

- The compiler injects the `srcDir` declaration into `app/build.gradle.kts` when it's missing (scaffolds installed before this change).
- It deletes legacy output: the old `src/main/java/com/nativephp/mobile/bridge/plugins/` dir wholesale, plus each installed plugin's package-derived copies under `src/main/java`, pruning only directories left empty — hand-written files sharing a package dir are untouched.

Known (unavoidable) gap: a stray copy from a file renamed or a plugin uninstalled *before* upgrading can't be located retroactively — `php artisan native:install` remains the escape hatch, and the limitation is documented in the code.

## Plugin compatibility

**No changes required from plugin authors.** The plugin contract is entirely package-based (bridge function FQCNs, renderers, init functions, relative component names, ProGuard keeps) and packages are unchanged; plugin repo layout is unchanged; the no-package fallback still works and still warns. Verified by the suite: every pre-existing plugin-behavior test passed with only copy-destination path literals updated. Only exotic exception: a hook that hardcodes a scaffold path like `app/src/main/java/<pkg>/…` to patch its own copied file would need to point at the new root — not a documented pattern; worth a release-notes line.

## Tests

New coverage: stale-copy pruning in the new root for the rename scenario (the original #188 bug), legacy-dir removal, legacy package-derived copy removal with empty-dir pruning, user code surviving next to a legacy copy, and idempotent `srcDir` injection (asserted inside the `android {}` block). Also smoke-tested the injection against an old-style scaffold `build.gradle.kts`.

Full suite: 614 passed / 4 risky / 3 skipped (risky/skipped pre-existing on `main`). Pint clean; PHPStan's only errors are pre-existing `JumpCommand` ones present on unmodified `main`.


shanerbaner82 (approved):
Approving — this is the right architecture (single compiler-owned root, wiped every compile, matching iOS) and the migration story is carefully done. Verified locally:

- Suite green on the branch as-is (614 passed) **and** merged with current `main` (868 passed = today's main baseline + this PR's 5 new tests) — the branch predates the Traits→Concerns rename but merges cleanly since it doesn't touch renamed files. Pint clean.
- Confirmed the wholesale `clean()` of the legacy `src/main/java/.../bridge/plugins` dir is safe: the template ships nothing else there — only the two registration placeholders, both moved in this PR. Core bridge files (`BridgeRouter.kt`, `PHPBridge.kt`, …) live one level up and are untouched.
- Checked the `removeLegacyPackageCopies` heuristic for the "user hand-wrote a same-named file in that package" worry — that state can't exist healthily, because the old compiler was already overwriting that exact path on every compile. Name+package match ⇒ compiler copy is sound.
- `pruneEmptyDirectories` passes `hidden: true` to `files()`, so a dir holding only a `.gitkeep` won't be pruned — nice touch.

One pre-existing oddity worth a follow-up, not a blocker: `PreparesBuild::replacePackagePlaceholder()` claims to rewrite `com.example.androidphp` in plugin files, but it only scans `src/main/java` — and since `prepareAndroidBuild()` runs *before* `compileAndroidPlugins()`, fresh plugin copies always overwrote any replacement anyway, so it was already ineffective for plugins under the old scheme (and no current first-party plugin uses the placeholder). This PR doesn't regress it, but the comment is now doubly stale — if the placeholder feature is meant to work for plugins, the substitution belongs in `copyPluginSources()`; otherwise the plugin claim in that docblock should go.

Tiny nit: the idempotency guard is `str_contains($content, 'src/nativephp/kotlin')`, so a user *commenting out* the injected `srcDir` line suppresses re-injection. Given they'd be deliberately disabling it, fine as-is.

PR #224, opened 2026-07-24
Readable labels for EDGE alignment (plus Concerns rename & comment cleanup)

simonhamp:
## Summary

This started from a question on X about why EDGE uses integers for alignment (`0=start`, `1=center`…) instead of strings. The reason is that alignment is a fixed-width integer in the binary shared-memory node format the native renderers read — but nothing stops us from adding a readable label layer on top. This PR does that, and bundles two related cleanups that came out of the same review.

## 1. Readable labels for EDGE alignment

Alignment now accepts an **enum**, a **string label**, or the **existing integer** — all normalize to the same wire value, so this is fully backward compatible.

- New int-backed enums under `Native\Mobile\Edge\Enums`: `AlignItems`, `AlignSelf`, `JustifyContent`, `TextAlign`, plus a shared `ResolvesAlignmentValue` trait.
- Fluent PHP: `->alignItems(AlignItems::Center)` (autocomplete + type-safety) or `->alignItems('center')`.
- Blade: `<native:column align-items="center" justify-content="space-between">`.
- `TailwindParser` (`items-*` / `justify-*` / `self-*` / `text-*`) now routes through the enums as the single source of truth.
- Removed the dead, never-referenced `Align` constants class — its `BASELINE = 4` was never a real native value (the native `FlexContainer` only supports align-items `0–3`).
- Added `AlignmentEnumsTest` covering the enums, `parse()`, the fluent setters, the Blade attribute path, and the Tailwind mapping.

Enum backing values match the native `FlexContainer` (iOS) / `ComposeFlexLayout` (Android) protocol exactly, so nothing changes on the wire.

## 2. `Traits` → `Concerns` (consistency)

Trait folders were a mix of `Traits/` and `Concerns/`. Standardized on `Concerns`, keeping Edge and root separate (no cross-area consolidation):

- `src/Traits` → `src/Concerns`, `src/Edge/Traits` → `src/Edge/Concerns`, `tests/Unit/Traits` → `tests/Unit/Concerns`.
- Namespaces and every `use` import updated; git preserves the rename history.

## 3. Yoga comment cleanup

Reworded code comments that implied NativePHP still uses Yoga to say "flexbox" (PHP plus the Swift/Kotlin readers). Layout is a native flex implementation — pure-Swift `Layout` on iOS, Compose `Layout` on Android. The **React Native comparison** references to Yoga (in the architecture doc and benchmark note) are accurate and were left intact.

## Docs

The nativephp.com docs (v4 `edge-components/layout.md`, `text.md`, `row.md`, `icon.md`) were updated to document the label/enum API. Those changes live in the `NativePHP/nativephp.com` repo and are **not** part of this PR.

## Verification

- Full test suite: **631 passing** (plus the new `AlignmentEnumsTest`).
- Pint: clean.
- PHPStan: unchanged vs the `origin/main` baseline — the 21 pre-existing larastan / Endroid-QrCode warnings are untouched by this branch (verified against a pristine `HEAD` checkout).


simonhamp:
Docs here: https://github.com/NativePHP/nativephp.com/pull/444

shanerbaner82:
Reviewed this one properly and liked it — so I've pushed two follow-ups straight to the branch rather than leaving nits. @simonhamp one of them is a deliberate behaviour change, so please sanity-check it before merge.

First, what I verified as-submitted:

- Wire values match native **exactly** — `AlignItems` 0–3 and `JustifyContent` 0–5 against both `FlexContainer.swift` and `ComposeFlexLayout.kt`, `TextAlign` 0/1/2 against `resolveTextAlignment`. Nothing moves on the wire.
- Dropping `Align::BASELINE` is safe — no `baseline` handling existed anywhere in the old `TailwindParser`.
- `Edge/Traits` → `Edge/Concerns` is a genuine straggler fix: `Edge/Concerns/` already existed (`Element` uses `Concerns\HasA11y`) and `Edge/Traits/` held exactly one file.
- 631 tests green, as claimed.

## 1. `parse()` now validates integers ⚠️ behaviour change

The stated reason for deleting `BASELINE = 4` is that native only supports 0–3 — but `parse()` forwarded any int untouched:

```
AlignItems::parse(4)  => 4
AlignItems::parse(99) => 99
AlignItems::parse(-1) => -1
```

All reached the wire, where the renderer hits its `default:` and silently falls back — which just moves the failure somewhere harder to find. Since we now have an enum, `tryFrom()` makes it a validation layer for free, and out-of-range values resolve to `null` so the setter leaves the native default in place.

**This replaces a documented pass-through** — the existing test asserted `parse(99) === 99` with the comment "Raw ints (existing call sites) pass through untouched — even unknown ones." I've updated that assertion, but if the pass-through was load-bearing for a call site I haven't spotted, say so and I'll revert this half. My read is that it directly contradicts the reasoning for removing `BASELINE`, so the two should agree.

## 2. Friendly aliases no longer leak into Tailwind classes

Routing `parseAlignItems`/`parseJustifyContent`/`parseAlignSelf` through `fromLabel()` pulled the whole alias set into the class parser. These all parsed on this branch and don't on `main`:

```
items-flex-start     => {"alignItems":0}
items-centre         => {"alignItems":1}
items-middle         => {"alignItems":1}
items-fill           => {"alignItems":3}
justify-spacebetween => {"justifyContent":3}
self-fill            => {"alignSelf":3}
```

None are real Tailwind classes. The aliases are genuinely nice for `align-items="centre"` in Blade and the fluent API — they just shouldn't define the *class* vocabulary, or we've forked the class language and we're exposed if Tailwind later gives those spellings a meaning.

So there's now a strict `fromUtilityClass()` alongside `fromLabel()`, and the utility parser uses it. `JustifyContent` overrides it because Tailwind spells those `justify-between`, not `justify-space-between`. `text-*` was already strict and is untouched.

Real Tailwind classes are **byte-identical** to before — I diffed the full `items-*` / `self-*` / `justify-*` / `text-*` matrix against the branch as submitted; no differences.

## Status

634 passing (631 + 3 new covering both fixes), Pint clean, PHPStan clean on the changed files.

## Not changed — your call

- **Invalid strings still fail silently.** `alignItems('bogus')` is a no-op where it used to be a `TypeError` from the `int` hint. That's an API design decision rather than a bug, so I left it — but it's the third place in EDGE where a typo'd value silently does nothing, so a `debugLog` on the unresolved branch might be worth a pass across all of them.
- **BC surface isn't recorded anywhere.** `Align` plus 26 `Native\Mobile\Traits\*` classes are removed/renamed. Third-party exposure is low (they're CLI internals), but there's no `UPGRADE.md` or `CHANGELOG.md` in the repo to note it, and stale trait imports in this ecosystem tend to fail in ways that look like unrelated boot crashes. Worth a line somewhere, or `class_alias` shims.

PR #226, opened 2026-07-24
feat(edge): stack z-overlay with two-point anchor/origin + attribute positioning

simonhamp:
## What

Two related positioning improvements to the Edge layout system.

### 1. Position via attributes (not just utility classes)

`buildLayoutArray` now accepts the short attribute spellings alongside the Tailwind classes:

```blade
{{-- these are equivalent --}}
<native:column class="absolute top-0 right-0" />
<native:column absolute top="0" right="0" />
<native:column position="absolute" :top="0" :right="0" />
```

- `absolute` / `relative` (bare boolean) and `position="absolute|relative"` → `position_type`
- bare `top` / `right` / `bottom` / `left` → the position offsets

The explicit camelCase form (`positionTop` from `top-4`) still wins on conflict. **No native change** — these produce byte-identical layout to the class forms.

### 2. `<native:stack>` is a real z-stack, with two-point anchoring

Previously `stack` fell through to a column-direction `FlexContainer` (aspirational per the docs, not implemented). Now its children are z-overlaid and absolutely positioned by default, and each child places itself with **two independent points**:

- **`anchor`** — the point on the **parent** the child hooks onto.
- **`origin`** — the point **on the child** that lands on that parent point.

Both default to `center`, and both are child attributes (or `anchor-*` / `origin-*` classes). Values: `center`, the four edges (`top`/`right`/`bottom`/`left`), and the four corners (`top-left`…`bottom-right`).

```blade
{{-- centered dot (default) --}}
<native:stack>
    <native:text class="text-[40] font-bold" font="pinkary">Pinkary</native:text>
    <native:column anchor="top-right" class="w-[14] h-[14] rounded-full bg-theme-primary" />
</native:stack>
{{-- ^ anchor=top-right, origin defaults to center → the dot's CENTER sits on the
     wordmark's top-right corner (drawing half outside — a plain stack doesn't clip). --}}

{{-- a badge whose bottom-left corner hooks the parent's top-right corner --}}
<native:column origin="bottom-left" anchor="top-right" ... />
```

Placement is `child.topLeft = parent.min + parentSize·anchor − childSize·origin` (+ insets as a nudge). The single-anchor case is just `origin == anchor`. The stack **sizes to its largest child** (ZStack/wrap-content), so no explicit size is needed. This lets a child draw **outside** its container — supported on both platforms as long as no clipping ancestor (rounded corners / scroll viewport) intervenes.

## How it's wired

`anchor` and `origin` ride the **tolerant props blob** (like `scroll_anchor`, `glass`) rather than the fixed binary layout struct (which is packed by an out-of-repo C extension and has no spare field) — so **no wire-format or C-extension change**. Shared enum: `center=0`, then the eight compass points `1..8`.

- **PHP** (`NativeElementCollector`, `TailwindParser`): resolves `anchor`/`origin`/`position`/short-offset attrs and `anchor-*`/`origin-*` classes; anchor/origin emitted through props on both the streaming and Element paths.
- **iOS** (`FlexContainer.swift`, `SwiftUINodeRenderer.swift`): a `stack` gets `isStack` — all children absolute, `sizeThatFits` = union of children, and `placeAbsolute` uses `W·anchor − w·origin` (+ insets).
- **Android** (`ComposeFlexLayout.kt`, `NodeView.kt`): a `stack` overlays children in a wrap-content `Box`, positioning each via a value-equal `TwoPointAlignment` built on Compose's `Alignment.align(size, space)`.

**Non-stack containers and non-anchored absolute children are byte-identical** — the new behavior is gated behind `isStack` / an explicit `anchor`|`origin`; plain `absolute top-0 right-0` keeps the legacy inset-corner behavior.

## Tests

26 PHP tests (`tests/Unit/Edge/LayoutPositioningTest.php`) — attribute↔class parity & precedence, the 9 anchor/origin names + aliases, independent anchor+origin on one element, anchor/origin ride props (never the layout struct). Full suite green (652), Pint clean.

## Needs on-device verification

Swift/Kotlin can't compile in the authoring environment. Please verify: a bare `<native:stack>` sizes to its child (not zero, not full-width); anchor+origin combinations land as expected (esp. a child drawing outside the container, and that hit-testing on the overflow behaves acceptably for interactive children); and the inset-as-nudge convention reads consistently across iOS/Android. Non-blocking: `app/src/main/baseline-prof.txt` pins the old `FlexContainer` signature — a stale generated-profile entry that just skips PGO for that method until regenerated.

### DIFF
--- a/resources/androidstudio/app/src/main/java/com/nativephp/mobile/ui/nativerender/ComposeFlexLayout.kt
+++ b/resources/androidstudio/app/src/main/java/com/nativephp/mobile/ui/nativerender/ComposeFlexLayout.kt
@@ -52,6 +52,41 @@ object Display {
 }
 
 /**
+ * 9-point anchor for `stack` container children (matches the PHP side).
+ * Read off each child's tolerant PROPS blob via `props.getInt("anchor")`;
+ * absent → CENTER (the stack default).
+ */
+object Anchor {
+    const val CENTER        = 0
+    const val TOP_LEFT      = 1
+    const val TOP_CENTER    = 2
+    const val TOP_RIGHT     = 3
+    const val CENTER_LEFT   = 4
+    const val CENTER_RIGHT  = 5
+    const val BOTTOM_LEFT   = 6
+    const val BOTTOM_CENTER = 7
+    const val BOTTOM_RIGHT  = 8
+}
+
+/**
+ * Maps an [Anchor] int to a Compose 2-D [Alignment]. Unknown / 0 → Center.
+ * All nine constants exist on `Alignment`'s companion (TopStart/TopCenter/
+ * TopEnd, CenterStart/Center/CenterEnd, BottomStart/BottomCenter/BottomEnd),
+ * so each maps 1:1 and can be handed straight to `BoxScope.align`.
+ */
+private fun anchorToAlignment(anchor: Int): Alignment = when (anchor) {
+    Anchor.TOP_LEFT      -> Alignment.TopStart
+    Anchor.TOP_CENTER    -> Alignment.TopCenter
+    Anchor.TOP_RIGHT     -> Alignment.TopEnd
+    Anchor.CENTER_LEFT   -> Alignment.CenterStart
+    Anchor.CENTER_RIGHT  -> Alignment.CenterEnd
+    Anchor.BOTTOM_LEFT   -> Alignment.BottomStart
+    Anchor.BOTTOM_CENTER -> Alignment.BottomCenter
+    Anchor.BOTTOM_RIGHT  -> Alignment.BottomEnd
+    else                 -> Alignment.Center // 0 CENTER (stack default) or unknown
+}
+
+/**
  * Renders children in a flex container using Compose's built-in Column/Row.
  * Maps flexbox properties to Compose equivalents:
  *   - flex_direction → Column or Row
@@ -72,8 +107,51 @@ fun FlexContainer(
     wrap: Int = 0,
     childNodes: List<NativeUINode>,
     modifier: Modifier = Modifier,
+    isStack: Boolean = false,
     content: @Composable () -> Unit
 ) {
+    // `stack` container: z-overlay EVERY child in a single Box, positioned by a
+    // 9-point `anchor` prop (default center) and fine-tuned by absolute inset
+    // offsets. No flow Column/Row runs — every child is treated as overlaid.
+    //
+    // Sizing: the Box carries only the incoming `modifier` (the stack node's own
+    // width/height, resolved upstream in NodeView). We add NO fillMaxWidth/
+    // Height/Size here, so a stack with no explicit size (WRAP mode → NodeView
+    // applies no size modifier) lets Box wrap-content and measure to its LARGEST
+    // child. `.align()`/`.offset()` are placement-only (ParentDataModifier /
+    // post-measure layout) and do not change what the Box measures, so the
+    // largest child still drives the stack's size — a text + a small anchored
+    // dot sizes to the text, never collapsing to the dot nor over-expanding.
+    if (isStack) {
+        Box(modifier = modifier) {
+            // Document order = z-order: first child drawn at the back.
+            childNodes.forEachIndexed { i, node ->
+                if ((node.layout?.display ?: 0) == Display.NONE) return@forEachIndexed
+
+                val alignment = anchorToAlignment(node.props.getInt("anchor", Anchor.CENTER))
+
+                // Inset offsets nudge the anchored child. Same convention as the
+                // absolute path: a set `right`/`bottom` pulls inward from that
+                // edge, otherwise `left`/`top` push from the leading edge.
+                val left = node.layout?.positionLeft ?: 0f
+                val top = node.layout?.positionTop ?: 0f
+                val right = node.layout?.positionRight ?: 0f
+                val bottom = node.layout?.positionBottom ?: 0f
+                val offsetX = if (right > 0f) (-right).dp else left.dp
+                val offsetY = if (bottom > 0f) (-bottom).dp else top.dp
+
+                Box(
+                    modifier = Modifier
+                        .align(alignment)
+                        .offset(x = offsetX, y = offsetY)
+                ) {
+                    NodeView(node = node)
+                }
+            }
+        }
+        return
+    }
+
     // Separate absolute children from flow children
     val hasAbsolute = childNodes.any { (it.layout?.positionType ?: 0) == PositionType.ABSOLUTE }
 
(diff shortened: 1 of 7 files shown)


shanerbaner82:
Built a demo screen covering every variation of this (anchor grid, origin grid, two-point combos, class↔attribute parity, inset nudges, legacy regression guards) and ran it on the iOS simulator with an instrumented build. Good news on the design; bad news on the wiring.

## The iOS stack path never executes

`SwiftUINodeRenderer.content` dispatches to `SwiftUIRendererRegistry` first, and `containerView` — the only place `isStack` is set — is the **fallback for unregistered types**. But `stack` *is* registered, in the generated `PluginRendererRegistration.swift`:

```swift
// Plugin: nativephp/mobile-ui
SwiftUIRendererRegistry.shared.register("stack") { AnyView(NativeUIStackRenderer(node: $0)) }
```

Runtime trace with `NSLog` in `makeCache`/`placeAbsolute`:

```
[ANCHOR] stack went to REGISTERED renderer — containerView/isStack skipped
[ANCHOR] makeCache isStack=false   ← every node, every container, whole app
```

55 types are registered by plugins, including `column`, `row`, `stack`, `scroll_view`, `pressable`, `canvas`. `containerView` is effectively dead for the entire real component surface. Visible symptom: all nine anchors render dead centre — `top-left` is pixel-identical to `bottom-right`.

**The placement math is correct**, for what it's worth — `column` does reach `FlexContainer`, and it lands exactly right:

```
place useTwoPoint=true anchor=3 bounds=(0,0,80,56) ideal=(18,18) -> (71.0, -9.0)
```

Top-right corner, half outside, as designed. It just never runs for stacks.

Android looks structurally identical — `NodeView.kt:49` hits `NativeRendererRegistry.get(node.type)` before the `DefaultContainerNode` fallback where `isStack` is set, and the registry's own docs say all core types are registered by the native-ui plugin. I verified the code path, not the runtime (only built iOS).

## `NativeUIStackLayout` is already the better implementation

This is the part that changes the fix. The existing stack layout isn't a naive ZStack — it carries four fixes, each documented with the bug it solved: children measured `.unspecified` so `maxWidth: .infinity` doesn't inflate them; scroll-views excluded from `maxSize`; `w-full` children measured at real width so wrapping text reports correct height; absolute children excluded from sizing. Its inset convention is `right != 0`, which supports negative insets (`-right-8` bleed) — the version here uses `right > 0`, which doesn't.

So this PR adds a second, competing stack implementation that is both unreachable and less capable than the one that ships.

**Correction to something I said earlier:** the "stack fills instead of wrapping" behaviour is real, but it's pre-existing in `NativeUIStackLayout.sizeThatFits` (`proposal.width ?? maxSize.width`) — not introduced here. Worth a separate issue.

## Suggested split

- **Drop the `isStack` branch** from `FlexContainer.swift` / `ComposeFlexLayout.kt`. It can't run, and it competes with a better implementation.
- **Move anchor/origin into `NativeUIStackLayout`** (and its Compose twin) in the `mobile-ui` repo — the math ports in ~15 lines. Which does mean this PR can't land its headline feature from this repo alone; core + plugin need to version together.
- **Keep the attribute-positioning half.** It works, verified on device — `class="absolute top-0 right-0"`, `absolute top="0" right="0"` and `position="absolute" :top :right` all land identically.

Two things to fix in that half:

1. `buildLayoutArray()` learned the short spellings but `applyLayout()` (the Element path) didn't, so they only work for the six builtin streaming types. `<native:column absolute top="0">` positions; `<native:text absolute top="0">` silently does nothing. Confirmed in the published tree — the `<text>` node comes out with padding only, no `position_type`, no `position`.
2. If you do unify those paths, watch out: `Circle` and `Rect` use `left`/`top` as their own drawing coordinates. They're safe today only because they go down the Element path — unifying would give every `<native:circle left="10" top="20">` phantom position insets.

Minor: `! empty($attrs['absolute'])` means `absolute="false"` sets absolute (non-empty string), and an unrecognised anchor name is dropped silently and centres — a `debugLog` on the `default => null` branch would save someone an afternoon.

## The broader thing

Worth pulling out separately: a plugin can silently take a node out of the layout system with no error, because layout policy is decided by renderer registration. React Native hit this in the old ViewManager architecture and fixed it structurally in Fabric — layout lives in the shadow tree, Yoga computes it for every node uniformly, and a host component can supply a measure function but never its own layout policy. Worth considering the same invariant here before the plugin surface grows further. Happy to open an issue with the longer version.

shanerbaner82 (changes requested):
This is a strong design — the two-point anchor/origin model is the right abstraction, the 9-point wire enum is identical across all three implementations (PHP `resolveAnchor`, Swift `anchorFractions`, Kotlin `Anchor`), riding the tolerant props blob avoids any wire-format change, and details like `TwoPointAlignment` being value-equal and deliberately LTR-absolute are well considered. Branch suite is green in its own era (652 passed) and the 26 new layout tests are good coverage.

Requesting changes for one structural reason: **#288 landed on `main` and rewrote the same code this PR rewrites, with a conflicting model** — so the rebase needs reconciliation, not just conflict resolution. I attempted the merge locally to map it: conflicts in `FlexContainer.swift`, `ComposeFlexLayout.kt`, and `NativeElementCollector.php` (the PHP ones are trivial prop-builder unions; the native ones are not).

**1. The "legacy inset" fallback is now a revert.** #288 replaced the old top-left/inset placement with real CSS semantics — one edge set anchors to it, both opposing edges stretch (`inset-0` fills), plus the signed-zero "authored zero" convention. This PR's non-anchor fallback path is a copy of the *pre-#288* behavior, so resolving the conflict in the PR's favor silently reverts #288 for every non-anchored absolute child. The fallback needs to become main's current `placeAbsolute` body on both platforms, with the anchor model layered on top (`isStack || anchor != nil || origin != nil` gate is fine as-is).

**2. Attribute-authored zeros break the equivalence claim — for the most common case.** On main, `top-0` emits IEEE **-0.0** (`TailwindParser::explicitInsets()`); +0.0 means "unset" and the native side tests `v != 0 || signbit`. The attribute path here casts `(float) $attrs['top']`, so `top="0"` produces +0.0 → *unset*. Post-merge, `<native:column absolute top="0" right="0" />` would not pin to the corner while `class="absolute top-0 right-0"` does — exactly the byte-identical promise in the PR description. The bare-attribute offsets need the same -0.0 marking for authored zeros (while the unset slots in the `[top, right, bottom, left]` array stay +0.0). Worth adding a zero-offset case to the parity tests — today they'd only exercise nonzero values, which is why nothing catches this.

**3. The nudge-vs-anchor inset semantics need a consistency statement.** After the rebase, `top-4` means *CSS edge anchor* on a plain absolute child but *positional nudge* on an anchored child. That's defensible, but the comments and PR text predate #288 and don't acknowledge the split — future readers should find the distinction stated where `placeAbsolute` branches.

Smaller, non-blocking:

- `<native:stack>` changing from fall-through column to z-overlay is the documented intent, but it's still a behavior change for any app that used it — worth a release-notes line.
- Invalid anchor names die silently: `anchor-typo` parses "successfully" in TailwindParser (so #296's new unsupported-class diagnostics won't flag it), then `resolveAnchor` returns null and drops it. A debug-mode warning would fit the diagnostics work that just landed.
- Your own open items still stand: on-device verification (stack sizing, overflow hit-testing, inset nudge parity) and the stale `baseline-prof.txt` entry.

Happy to pair on the reconciliation if useful — the merge map above is already done locally.

PR #236, opened 2026-07-26
Docs/icons enums path

martin-ro:
Core Boost guidelines and skills currently tell agents to run `native-ui:generate-icons` and import `App\Icons\*`. Neither belongs here.

Core owns `<native:icon>`, the name resolution in `src/Icon/IconResolver.php`, the `:ios` / `:android` binding in `src/Concerns/HasPlatformIcon.php`, and the `IosSymbol` / `AndroidSymbol` interfaces. It does not own the enum catalogs. Those come from `nativephp/mobile-ui`, which requires core but is not required or suggested by it, so an app running core alone gets told to run a command it doesn't have.

This drops the generation specifics (the command, `--namespace` / `--output`, `--refresh-material`, the namespace, and the `@use` example) and points at mobile-ui's guidelines instead. What core actually backs stays: prefer typed enums, bind them with `:ios` / `:android`.

Also in here: `App\Icons` becomes `App\Enums\Icons` in `references/available-docs.md`, and a `HasPlatformIcon` docblock that named a `Nativephp\NativeUi\Icon` namespace that never existed is corrected.

`nativephp-v3-to-v4-upgrade/SKILL.md` still names the command in its phase 2 ordering. Left alone on purpose, since that skill walks someone through adopting native-ui and states no namespace. The branch also carries `11b4878 Cleanup duplication`, which is unrelated and a no-op against current main.

Depends on the [mobile-ui PR that adds the guidance](https://github.com/NativePHP/mobile-ui/pull/25) this now points at. Merge that one first.


shanerbaner82:
@martin-ro the reasoning here is sound and the underlying problem is real — core was telling agents to run a command it doesn't ship. I checked the claims rather than taking them on trust, and they all hold:

- **Core doesn't own `native-ui:generate-icons`** — zero references in `src/`; it's mobile-ui's `src/Console/GenerateIconsCommand.php`.
- **Core neither requires nor suggests mobile-ui** — confirmed in `composer.json`. So an app on core alone really was being pointed at a command it doesn't have.
- **The enums are generated, not shipped** — no `Ios.php` / `Android.php` / `AndroidOutlined.php` anywhere in mobile-ui's `src/`.
- **`Nativephp\NativeUi\Icon\*` never existed** — `git log -S` across all of mobile-ui's history finds nothing, though `namespace Nativephp\NativeUi` itself did exist in three commits. So the parent was real and the `Icon` segment never was. That docblock was wrong twice over: wrong namespace *and* "plugin-supplied" when they're generated into the app. Good catch.
- **The `App\Enums\Icons` change is correct but strictly dependent** — the command still defaults to `app/Icons` today; mobile-ui#25 is what moves it. Your merge-order note is load-bearing: landed first, this makes core's docs wrong in a new way rather than fixing them.

Two things to fix before it goes in, though.

## 1. `nativephp-mobile/SKILL.md` contradicts itself now

You removed the `@use` lines from the icon section and replaced them with "see its guidelines for generating and importing them" — but the same file still has them at lines 198–199, in the top-bar chrome example:

```blade
@use('App\Icons\Ios')
@use('App\Icons\Android')
```

This is the one I'd most want fixed, because an agent will copy a working code block over prose every time. It also carries the pre-#25 namespace, so it goes stale the moment the dependency lands — and it's in a file this PR already touches, so the inconsistency is self-inflicted.

## 2. The rationale for skipping the upgrade skill doesn't hold

The description says `nativephp-v3-to-v4-upgrade/SKILL.md` was left alone because it "states no namespace". Line 149 does:

```
- **Icons**: `App\Icons` enums via `:ios`/`:android`; a raw SF Symbol / Material name string is
  accepted when an enum case is missing.
```

The `generate-icons` mention at line 81 is genuinely fine to leave — that skill is explicitly about adopting native-ui, so the command *is* in context there, and I agree with your call on it. But line 149 pins a namespace that #25 invalidates. Either update it to `App\Enums\Icons` or drop the namespace entirely and say "the generated icon enums", which is the more robust option given it's the one thing that keeps moving.

## Minor

Core's guidelines now point at "the `nativephp/mobile-ui` package; see its guidelines" for a package core neither requires nor suggests. Strictly better than naming a command that doesn't exist, but an agent following it still dead-ends when mobile-ui isn't installed. A `composer require nativephp/mobile-ui` in that sentence would make the pointer actionable.

Happy to push the two `App\Icons` fixes to your branch if that's easier — maintainer edits are on. Otherwise this is an approve from me once they're in and mobile-ui#25 has landed.

martin-ro:
@shanerbaner82 Both fixes are in (81dd90e), plus your minor suggestion:

- The top-bar chrome example at L198-199 now imports `App\Enums\Icons\*` — the one spot that stays namespace-pinned, since a runnable code block needs a concrete path.
- The v3-to-v4 upgrade skill L149 drops the namespace entirely ("the generated icon enums"), per your more-robust option. L81 left as-is, as agreed.
- Core guidelines now include `composer require nativephp/mobile-ui` so the pointer is actionable.

Merge order noted — this waits on mobile-ui#25.

simonhamp:
I kinda prefer that these enums just live in core or mobile-ui tbh (leaning more towards core) - I'm not clear on why we should be making the user build the enums in their app.

I do think that there probably does need to be a generator command as the icons do get updated with new OS releases, but that wouldn't be used by devs, more just in our CI/release process to ensure we have a full set of icons appropriately represented by the enums.

PR #237, opened 2026-07-26
Rebuild iOS toasts: custom views, dismissal, timing, position and a visible stack

simonhamp:
iOS toasts were a fixed black label at the bottom of the screen, shown one at a time from a queue: nothing about them was under the developer's control, and a burst of toasts left the user staring at the first one with no sign that more were waiting.

## Visible stack

Toasts are now presented in a stack inside a passthrough overlay window (touches that miss a toast fall straight through to the WebView).

- The front toast is the one being read; new toasts **grow into place behind it**, scaled down slightly and peeking out, while the stack nudges a few points towards its edge — so it's obvious more is coming.
- The stack holds a maximum of **5** toasts. Anything pushed while it's full is **silently dropped**, never queued.
- **Only the front toast counts down**, so nothing disappears before it's been read. When it goes, the next comes forward and starts its own timer.
- A stack always stays where it started: a toast asking for the opposite position joins the existing stack rather than splitting it in two.

## Developer control

- **Timing** — `duration()` in seconds, `short()` / `long()`, or `persistent()` to stay until dismissed. Omitted keeps the existing reading-speed calculation.
- **Location** — `top()` / `bottom()` (a hint, per the stacking rule above).
- **Dismissal** — tap, or swipe towards the anchored edge or sideways, with rubber-banding and fling detection. Only the front toast is interactive; `dismissible(false)` toasts don't take touches at all and are dismissed programmatically via `Toast::dismiss($id)`.
- **Appearance** — `view()` renders any view as the toast body.

View-backed toasts render in a transparent, non-scrolling `WKWebView` using the `php://` scheme handler and the app's data store, loaded against the app's own base URL, so relative asset paths (compiled CSS, images, fonts) resolve exactly as they do everywhere else. Height is measured from the rendered document and clamped to 60% of the screen; a slow render is shown at a fallback height and resized when it settles, and a failed base-URL load retries once without it rather than vanishing.

## PHP API

```php
use Native\Mobile\Facades\Toast;

Toast::message('Saved');                                   // auto-shows, like PendingAlert
Toast::message('Saving…')->top()->persistent()->id('saving');
Toast::view('toasts.order-shipped', ['order' => $order])->duration(6);
Toast::html('<div class="toast">Saved</div>')->short();

Toast::dismiss('saving');
Toast::dismissAll();
```

`view()` accepts a view name, a `View` instance or any `Htmlable`. Fragments are wrapped in a minimal transparent document; full documents are left alone.

## Changes

- `resources/xcode/NativePHP/Components/Toast.swift` — rewritten: `ToastManager` (stack, positioning, timing, overlay window), `ToastView` (label or web-view content, gestures, measurement), `ToastWindow` (passthrough hit testing).
- `resources/xcode/NativePHP/Bridge/Functions/ToastFunctions.swift` — new `Toast.Show`, `Toast.Dismiss`, `Toast.DismissAll` bridge functions, registered in `BridgeFunctionRegistration.swift`. The Xcode project uses synchronized file groups, so the new file needs no `.pbxproj` change.
- `src/Toast.php`, `src/PendingToast.php`, `src/Facades/Toast.php` — the new PHP API.
- `src/Dialog.php` — `Dialog::toast()` is deprecated and delegates to the new API.
- `resources/dist/native.js` — matching `Toast` builder for JS/Inertia apps.
- `tests/Unit/PendingToastTest.php` — covers defaults, positions, durations, dismissibility, view/HTML rendering and the guards.

## Compatibility

- `Dialog::toast()` and the global `showToast(message:duration:)` helper the dialog plugin calls both keep working.
- Platforms without the `Toast.*` bridge functions fall back to `Dialog.Toast` with the plain text message, so **Android behaviour is unchanged**. A view-only toast is skipped there rather than shown as raw markup.

## Testing notes

CI runs the suite for this branch. In the authoring sandbox `composer install` was blocked by the network proxy, so Pest/Pint/PHPStan didn't run there; the PHP builder logic was verified through a stub harness plus `php -l`, and `native.js` with `node --check`. The Swift is unbuilt — it needs a simulator/device pass to confirm the stack animation, gesture dismissal and view-backed sizing look right.

https://claude.ai/code/session_01JiHVqJqoQ1ased6Y6TyJNn

---
_Generated by [Claude Code](https://claude.ai/code/session_01JiHVqJqoQ1ased6Y6TyJNn)_


shanerbaner82:
@simonhamp ran everything that couldn't run in your sandbox — **760 tests pass, Pint clean, PHPStan clean** on the new PHP. The builder is careful work: `toArray()` filtering on `!== null` rather than truthiness is what keeps `duration: 0` and `dismissible: false` alive, and the destructor guarding on "has content" before calling `show()` neatly sidesteps the throwing-destructor trap.

I've pushed one fix (`ad3e1c0`), and there's a follow-up in Jump that I think should gate the merge.

## The bug: Jump + Android silently loses every toast

`supportsStackedToasts()` gates on `nativephp_can('Toast.Show')`. Under Jump that function is a stub:

```php
// In Jump hybrid mode, we assume all functions are available
// on the connected device.
function nativephp_can(string $method): bool
{
    return true;
}
```

`Toast.Show` is registered only in `BridgeFunctionRegistration.swift` — Android has nothing but `Dialog.Toast`. So Jump against an Android device sends `Toast.Show` into the void.

And since `Dialog::toast()` now delegates to the new API, **every existing call goes with it**. That's the opposite of "Android behaviour is unchanged" — which does hold on-device, just not under Jump.

**Fixed by gating on `Platform::isAndroid()`.** `Platform::current()` probes the real device over the bridge, so unlike `nativephp_can()` it answers honestly in both environments. The capability check still runs afterwards, so an older iOS runtime without `Toast.*` keeps falling back rather than dispatching into the void.

I deliberately bail only on a **known** Android device — an unknown platform (no bridge, tests) keeps the previous behaviour, so this can't lose toasts anywhere new. Three regression tests; the first two fail if you remove the gate (I checked).

## The merge should be contingent on a Jump fix too

My change fixes toasts. It doesn't fix the underlying problem: **`nativephp_can()` lying under Jump.** Every future bridge function that isn't on both platforms hits this exact wall, and the failure mode is always the same — silent, and only in Jump, which is where most day-to-day development happens.

Right now the only defence is that each caller remembers to add a platform check by hand, which is precisely the kind of thing that gets forgotten in the PR that introduces the third iOS-only function.

Worth doing properly: have Jump's `nativephp_can()` ask the connected device rather than assuming. The device already has a real registry to answer from, and Jump already has a bridge to ask over — `Platform::current()` does an equivalent round-trip and pays for it with a 2s throttle on failure, so there's a working pattern to copy. Could you pick that up as part of this, or as an immediate follow-up? I'd rather not merge an iOS-only bridge function while the capability check that's supposed to guard it can't see platforms.

## Smaller things, none blocking

- **`PendingToast::MAX_STACK` is dead code.** Declared and documented, never referenced; Swift keeps its own `ToastMetrics.maxStack = 5`. Two constants that have to agree with nothing enforcing it — same drift shape as the `PropKey` table mirrored across three files. Delete it or use it.
- **`persistent()` degrades oddly on Android:** `duration(0)` → `0 <= DURATION_SHORT` → `'short'`, so a toast meant to stay until dismissed shows for 2s, and `Toast::dismiss($id)` has nothing to dismiss. Worth a line in the docs.
- **`Str::contains($content, '<html')`** treats any fragment merely mentioning that literal (a code sample, escaped example markup) as a full document.

Credit where it's due: the stack-full drop **is** logged natively (`print("🍞 Toast dropped, stack is full…")`), so it isn't a silent drop.

## Still unverified

The Swift is unbuilt, as you say. Beyond the animation and gestures, the piece I'd most want eyes on is the view-backed toast — a `WKWebView` per toast on the `php://` scheme, height measured from the rendered document, a fallback height for slow renders, and a retry-without-base-URL path. That's a lot of failure modes, and up to five can be live at once.

When you build it: `native:install` before `native:run`, or the new Swift won't reach the binary. `native:run` recompiles the existing `nativephp/ios` scaffold and doesn't re-copy from the package — cost me two builds on #226 before I spotted it.

simonhamp:
Thanks — the Jump + Android catch is a good one, and `Platform::isAndroid()` is the right gate. Your `Platform::set()` cleanup is better than the reflection helper I'd used, and it caught my test that was still calling the old one.

The three smaller items are fixed in `362c3cf`:

- **`MAX_STACK`** deleted. The limit lives in `ToastMetrics.maxStack`; the stack behaviour is described on the `PendingToast` class docblock instead, so there's no second copy to drift.
- **`persistent()` on Android** now falls back to `'long'` rather than `'short'` — `fallbackDuration()` treats "no duration" and "until dismissed" alike as *the longest we have*, which is as close as Android's API gets. `persistent()` says so in its docblock, including that `Toast::dismiss()` has nothing to dismiss there. Covered by a test.
- **Document detection** is anchored: `/^(<!doctype\s+html|<html[\s>])/i` against the left-trimmed content. A fragment that merely mentions `<html>` in a code sample is a fragment and gets wrapped; only something that actually starts as a document is passed through. Also covered.

## On `nativephp_can()` under Jump

Agreed on the diagnosis, and we're not fixing it in this PR — here's the constraint, since it changes what "fix it properly" costs.

The device side of Jump isn't in this repo. `resources/jump/websocket-server.php` is the dev-machine relay: it forwards `bridge_call` frames and has no opinion about what's registered. The thing that could answer a capability question is the Jump app, which is a released binary. So "ask the connected device" needs three things, in order:

1. a responder — something like `Bridge.Can` registered in `BridgeFunctionRegistration.swift` *and* `BridgeFunctionRegistration.kt`, since a check that only answers on one platform reintroduces the same blind spot;
2. `nativephp_can()` in the Jump polyfill querying it, and **still** returning `true` when the device doesn't recognise the query — otherwise every Jump app older than that release starts silently dropping calls, which is the current bug with the sign flipped;
3. a Jump app release built from those templates, before any of it takes effect.

Until step 3 ships, PHP has to fail open, so the platform gate in this PR is doing the real work either way. Until someone picks the capability work up, an iOS-only bridge function needs its own gate — as here.

Two notes on the review itself: the `duration: 0` / `dismissible: false` survival is exactly what the `!== null` filter is for, so that read is right; and re: `native:install` before `native:run` — noted for the simulator pass on the view-backed toast, which is the part I'd most want eyes on too.

---
_Generated by [Claude Code](https://claude.ai/code)_

simonhamp:
Refreshed this against `main` (it was 76 commits behind) — merged clean, no conflicts. Two things worth flagging before anyone builds it for the `needs testing` pass.

## `UI.SetBackground` paints the toast window

`UI.SetBackground` landed while this branch was open, and it paints every window in the scene except UIKit's keyboard helpers:

```swift
windowScene.windows.filter { window in
    let className = String(describing: type(of: window))
    return !className.contains("TextEffects") && !className.contains("RemoteKeyboard")
}
```

The toast stack adds a window of its own — a transparent passthrough `ToastWindow` that exists for as long as a toast is on screen. Painting it produces the exact failure the keyboard-helper exclusion was added for in #373/#393: an opaque sheet of the background colour over the whole app, with the app still rendering and taking taps underneath, for the duration of the toast.

It only bites when `SetBackground` runs *while* a toast is up (the window is created lazily and torn down when the stack empties), which is why both features can pass on their own and still be broken together. Fixed in `3b7f06e` by excluding it in `appWindows` — by type, since this window is ours, rather than by class-name prefix.

Worth a look during the device pass: show a toast, then call `UI.SetBackground` from PHP while it's visible.

## Still unverified

Unchanged from before — the Swift has never been compiled or run. The view-backed toast is the part carrying the most risk: a `WKWebView` per toast on the `php://` scheme, height measured from the rendered document, a fallback height for slow renders, and a retry-without-base-URL path, with up to five live at once.

One merge-related note for whoever builds it: `WebView.dataStore` on `main` is now `WKWebsiteDataStore.default()` rather than `.nonPersistent()`. Toast web views share it, which is what we want for session cookies on asset requests — just no longer the same store this branch was written against.

---
_Generated by [Claude Code](https://claude.ai/code)_

simonhamp:
Given that someone has built a package for [improved toasts](https://github.com/victorycodedev/toastkit), I'm going to close this. And I think we should probably deprecate the built-in toast functionality, removing it in the next major release

PR #238, opened 2026-07-26
Typed platform values and a capability seam for tests

shanerbaner82:
Two testability gaps that let a real regression through in #237.

## `Platform::set()` accepted any string

```php
Platform::set('androdi');   // accepted silently
Platform::isAndroid();      // false, for the rest of the process
```

A test seam you can't trust is worse than not having one. It now takes an `Os` enum case or its string spelling, and throws on anything else:

```php
Platform::set(Os::Android);   // typed
Platform::set('ios');         // still works
Platform::set('iOS');         // casing tolerated
Platform::set('androdi');     // InvalidArgumentException
```

`Os` is backed by the same `'ios'` / `'android'` values `Platform` has always used, so **nothing changes on the wire and no existing call site changes**. `current()` still returns `?string`; `Platform::IOS` / `ANDROID` are now expressed in terms of the enum so they can't drift. New code can use `Platform::os()` for a value that can't be compared against a typo.

Named `Os` rather than `Platform` because `Native\Mobile\Platform` is the class that resolves it, and two `Platform` symbols in adjacent namespaces would be a permanent source of wrong imports.

## `nativephp_can()` had no seam at all

`FakeBridge` intercepts `call()` and the element functions but never `nativephp_can()`, and the polyfill hardcodes `return true`. So **the fallback arm of every capability-gated branch was unreachable from the suite** — it could only be reached indirectly, if at all.

That is how an iOS-only `Toast.Show` reached review silently swallowing every Android toast, including existing `Dialog::toast()` calls (#237). The guard was there; nothing could test what happened when it said no.

```php
$bridge = FakeBridge::enable()->cannot('Toast.Show');

Toast::message('Saved')->show();

$bridge->assertCalled('Dialog.Toast')
       ->assertNotCalled('Toast.Show')
       ->assertCapabilityChecked('Toast.Show');
```

Everything stays available by default, so existing tests are unaffected.

## Deliberately out of scope

The Jump polyfill still answers `true` to every capability probe regardless of the connected device:

```php
// In Jump hybrid mode, we assume all functions are available
// on the connected device.
function nativephp_can(string $method): bool
{
    return true;
}
```

That's the behavioural half of the same problem and it's being picked up on #237, so I've left it alone rather than have two PRs edit the same function. This PR is the testing half — it gives that work something to assert against.

## Tests

10 new, covering enum and string input, casing, the rejection, the null reset, constant/enum agreement, and the capability seam in both directions. Full suite **748 passing**, Pint clean, PHPStan clean on all three changed files.

### DIFF
--- a/src/Enums/Os.php
+++ b/src/Enums/Os.php
@@ -1,0 +1,44 @@
+<?php
+
+namespace Native\Mobile\Enums;
+
+use Native\Mobile\Platform;
+
+/**
+ * The mobile operating systems NativePHP targets.
+ *
+ * Backed by the same lowercase strings {@see Platform} has
+ * always used on the wire (`'ios'` / `'android'`), so this is a typed
+ * spelling of the existing values rather than a new vocabulary — anywhere a
+ * string is accepted today, an enum case now is too.
+ *
+ * Named `Os` rather than `Platform` because `Native\Mobile\Platform` is the
+ * class that resolves it, and two `Platform` symbols in adjacent namespaces
+ * would be a permanent source of wrong imports.
+ */
+enum Os: string
+{
+    case Ios = 'ios';
+
+    case Android = 'android';
+
+    /**
+     * Resolve a case-insensitive name, or null when it isn't one of ours.
+     * Unlike `tryFrom()` this tolerates the casing people actually write
+     * (`iOS`, `Android`).
+     */
+    public static function fromLabel(string $label): ?self
+    {
+        return self::tryFrom(strtolower(trim($label)));
+    }
+
+    public function isIos(): bool
+    {
+        return $this === self::Ios;
+    }
+
+    public function isAndroid(): bool
+    {
+        return $this === self::Android;
+    }
+}
--- a/src/Platform.php
+++ b/src/Platform.php
@@ -2,6 +2,7 @@
 
 namespace Native\Mobile;
 
+use Native\Mobile\Enums\Os;
 use Native\Mobile\Facades\System;
 
 /**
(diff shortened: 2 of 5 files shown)


simonhamp (commented):
Honestly this all just feels like massive overkill - imo no one should be able to `set` the platform - the platform defines that itself.

We already have [`System`](https://github.com/NativePHP/mobile-air/blob/main/src/System.php)

I'm not sure what `Platform` brings to the table.

tbf, even `System` could be simplified... right now it makes a full native call to get the platform, but if each shell app just set an explicit string in the environment of the PHP runtime at startup (e.g. `setenv('PLATFORM=ios')` and `setenv('PLATFORM=android)`) then we wouldn't need any of this `Platform::set()` stuff

This would be testable then just by setting the appropriate env key in the test environment or test setup.

This would then be very simple (and fast) string comparison in a very simplified convenience class (probably System, where it already is)

shanerbaner82:
Pretty sure we already do exactly that — both shells set `NATIVEPHP_PLATFORM` at boot (`NativePHPApp.swift` / `PersistentPHPRuntime.swift` on iOS, `LaravelEnvironment.kt` on Android) and it already lands in `nativephp-internal.platform` config. `Platform` just isn't reading it — it probes the bridge instead.

So happy to rework this to read the env value, and all the retry/throttle stuff disappears on device. The one wrinkle is Jump: the shell doesn't launch PHP there, so the env isn't set and the platform only arrives once the device attaches — that case still needs the bridge probe (or the handshake could set it per-connection).

Cool if I split the `FakeBridge` capability seam into its own PR? That half is just about making `nativephp_can()` fallbacks testable (the gap that let #237 through) and doesn't depend on how platform detection works.

PR #246, opened 2026-07-30
Create storage/framework runtime dirs before composer install

shanerbaner82:
Fixes #245

## Problem

`native:run` fails on a fresh app during "Installing Composer dependencies" with:

```
InvalidArgumentException
Please provide a valid cache path.
at vendor/laravel/framework/src/Illuminate/View/Compilers/Compiler.php:75
```

Bundle prep excludes `storage/framework` from the copy (`BundleExclusions::PROJECT`) but only recreated `bootstrap/cache` before running `composer install`. Composer's `post-autoload-dump` runs `package:discover`, which boots the app in the temp tree; the Blade compiler resolves `view.compiled` via `realpath(storage_path('framework/views'))`, which returns `false` for the missing directory. The empty `storage/framework/{cache,sessions,views}` dirs were only added later, in `createZipBundle()` — after Composer.

## Fix

Ensure `storage/framework/{cache,sessions,views}` alongside `bootstrap/cache` before `composer install` runs, exactly as suggested in the issue. The list was already duplicated in both `createZipBundle()` branches (ZipArchive and 7-Zip), so it's consolidated into a single `RUNTIME_DIRS` constant used by all three sites.

## Test

New regression test in `ReleaseBuildBundleTest` uses a `Process::fake` closure — evaluated at install time — to observe the temp tree exactly as `composer install` would. It fails without the fix (all three `storage/framework` dirs missing) and passes with it. Full suite: 788 passed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #247, opened 2026-07-30
Recreate the runtime directories Laravel needs in the bundle copy

gwleuverink:
`native:run` stops at `composer install` with `Please provide a valid cache path`, before a build starts. It hits any app that has not published `config/view.php`, which is every fresh skeleton.

`BundleFileManager::copy()` excludes `storage/framework`, then composer runs `package:discover` in the copied tree, which boots Laravel. Without a published view config the framework default applies, and it wraps the path in `realpath()`. That returns false for a directory that is not there, and the Blade compiler refuses a false path.

The zip copy this replaced already handled it: it excluded the same contents, then added `bootstrap/cache`, `storage/framework/cache`, `storage/framework/sessions` and `storage/framework/views` back as empty entries. Consolidating onto `BundleFileManager` carried the exclusions across but not that restoration. Android kept `bootstrap/cache` by hand, iOS kept none.

This adds `BundleExclusions::REQUIRED_DIRECTORIES` and recreates them at the end of the copy, so the exclusion set and the carve-out it makes necessary stay in one place and every caller gets both.

Test fails without the change and passes with it. Verified on iOS & Android emulators.

### DIFF
--- a/src/Support/BundleExclusions.php
+++ b/src/Support/BundleExclusions.php
@@ -50,6 +50,27 @@ class BundleExclusions
         'artisan',
     ];
 
+    /**
+     * Working directories recreated after the exclusions above have run.
+     *
+     * Their contents are excluded on purpose, being compiled output and one
+     * machine's sessions, but the directories themselves are not optional:
+     * composer install runs package:discover in the copied tree, and that
+     * boots Laravel.
+     *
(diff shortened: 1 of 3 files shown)


gwleuverink:
Duplicate of #246

PR #248, opened 2026-07-30
feat(edge): dispatch screen lifecycle events from the navigation loop

sylvesterdamgaard:
## The problem

Cross-cutting observers — telemetry, analytics, crash breadcrumbs — have nowhere to attach to the screen lifecycle today.

`NativeRouter` constructs components with `new $class`, the router itself is constructed directly rather than resolved from the container, and no events are dispatched. So the only way in is to override `mount()` / `onResume()` / `unmount()` from a shared base class — and that competes with the app for those methods. A screen that defines its own `mount()` silently replaces the instrumented one, and the observer goes quiet on exactly the screens someone bothered to write logic for. An observer you can't trust to be complete isn't much use.

## The change

Dispatch `ScreenMounted` / `ScreenResumed` / `ScreenUnmounted` from the navigation loop, *alongside* the component's own hook rather than in place of it. Nothing about the existing lifecycle changes.

- The seven `unmount()` call sites in `loop()` now route through one `unmountComponent()` helper, so the event has a single home rather than seven.
- Listener exceptions are caught and sent to `debugLog()`. An observer must not be able to take the navigation loop down with it.
- `announce()` guards on `function_exists('event')`, so the router stays usable outside a booted container — which the testing helpers rely on.
- Preloaded stacks (hot restart) deliberately stay quiet: those screens are being restored, not navigated to. Happy to change that if you'd rather they fire.

Events carry the component class and the uri, and use plain `Dispatchable` — not `BroadcastsGlobally`, since these originate in PHP rather than from the native side.

```php
Event::listen(ScreenMounted::class, fn ($e) => ...); // $e->component, $e->uri
```

## Verification

Ran against `main` on PHP 8.4:

- test suite **787 → 789 passed** (the two added here), same 4 risky / 3 skipped as the baseline
- `phpstan` clean

## Context

Found while adding NativePHP support to [cboxdk/laravel-telemetry](https://github.com/cboxdk/laravel-telemetry). The runloop holding a single request open for the lifetime of a screen means `Kernel::terminate()` never fires, so anything built around the request lifecycle needs different anchor points — these events are the missing ones. That package works without this change, but the mount/unmount half of the instrumentation has to be left out to stay honest.

Open to a different shape here — naming, payload, whether the router is the right place to dispatch from. Happy to rework.

### DIFF
--- a/src/Edge/NativeRouter.php
+++ b/src/Edge/NativeRouter.php
@@ -2,6 +2,10 @@
 
 namespace Native\Mobile\Edge;
 
+use Native\Mobile\Events\Screen\ScreenMounted;
+use Native\Mobile\Events\Screen\ScreenResumed;
+use Native\Mobile\Events\Screen\ScreenUnmounted;
+
 class NativeRouter
 {
     /**
@@ -423,9 +427,11 @@ class NativeRouter
                     $component->publishPlaceholder();
                     static::debugLog('loop: calling mount() on '.get_class($component));
                     $component->mount();
+                    $this->announce(new ScreenMounted(get_class($component), $entry['uri'] ?? null));
                 } else {
                     static::debugLog('loop: calling onResume() on '.get_class($component));
                     $component->onResume();
+                    $this->announce(new ScreenResumed(get_class($component), $entry['uri'] ?? null));
                 }
             } catch (NativeDumpException $e) {
                 $component->renderDumpScreen($e);
@@ -448,7 +454,7 @@ class NativeRouter
 
             if ($intent === null) {
                 static::debugLog('loop: no intent, popping '.get_class($component));
-                $component->unmount();
+                $this->unmountComponent($component);
                 array_pop($this->stack);
                 $freshPush = false;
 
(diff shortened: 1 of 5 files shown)


simonhamp (approved):
Love it! Thanks @sylvesterdamgaard

simonhamp:
@sylvesterdamgaard would you adding a PR for the docs about these new events?

sylvesterdamgaard:
> @sylvesterdamgaard would you adding a PR for the docs about these new events?

Sure thing. I will get that fixed now. :)

sylvesterdamgaard:
Docs are up: NativePHP/nativephp.com#452.

New **Observing the lifecycle from outside** section on the Lifecycle Hooks page — when each of the three events fires, what they carry, and how to listen — plus a pointer from the Events page, which so far only covered component-targeted native events.

I also documented the two behaviours that aren't obvious without reading the router: a listener that throws is logged rather than propagated, and restoring a preloaded stack after a hot reload stays quiet.

It describes events that aren't merged yet, so hold or merge it in whatever order suits you.

simonhamp:
Thanks so much!

PR #253, opened 2026-08-02
Fix native route middleware during in-app navigation

### LINKED ISSUE #252: v4: Route::native() accepts ->middleware() but only applies it on cold start

`Route::native()` returns a real `Route::get()`, so `->middleware()` chains onto it and looks like it works. It runs for the screen the app launches into, because the runloop is entered inside that HTTP request. It does not run for any screen reached by in-app navigation, which resolves through `NativeRouter::resolve()` and mounts the component directly with no request and no pipeline.

The scope is wider than auth. Routes registered from `withRouting(then: …)` land in no route group at all, so they carry no middleware whatsoever - not even web. php artisan route:list -v shows our native routes bare, next to Livewire’s showing web.

### What happened?

I hit this migrating an app from v3 to v4. The v3 routes were:

``
Route::middleware(['auth.device', 'auth.lock'])->group(function () {
    Route::livewire('/', 'pages::dashboard');
    // …13 more
});
``

Moving them to `Route::native()` silently dropped every guard. Nothing failed, nothing logged. Our test suite couldn’t catch it either - component tests mount screens directly. 

### How to reproduce the bug

See above 

### Debug Output

See above 

_No response_

Jump app (iOS)

### Notes

_No response_

wojt-janowski:
Closes #252.

## Problem

`Route::native()` returns a real Laravel route, so chained middleware runs when the screen is entered through the initial HTTP request. In-app navigation takes a different path: `NativeRouter` resolves the URI and mounts the component directly for push, replace, and hot-reload stack restoration.

That means route middleware such as `auth`, `can`, and custom guards can be silently skipped after the app starts. A protected native screen can therefore mount when reached from another native screen even though the same URL is guarded on cold start.

## Fix

- Retain the Laravel route associated with each native route without changing the public native-route registry shape.
- Before in-app push, replace, or stack preload, clone and bind that route to a synthetic request for the destination URI.
- Resolve aliases, groups, exclusions, and priority through Laravel's real route middleware resolver.
- Scope the request, route binding, Route facade, and router request context while running Laravel's routing pipeline, then restore the long-lived context.
- Return blocked middleware responses unchanged, preserving redirect status, cookies, headers, and non-redirect responses.
- Skip protected historical entries during hot-reload stack restoration.
- Leave the cold-start HTTP path unchanged so middleware is not run twice.

The stored route is cloned before parameter binding, preventing navigation between values of the same route pattern from mutating earlier route state.

## Verification

- `php84 vendor/bin/pest` — 800 passed, 2,799 assertions
- `php84 vendor/bin/phpstan analyse --no-progress` — no errors
- Pint and `git diff --check` pass
- iOS simulator reproduction:
  - released `4.0.1`: in-app navigation mounted the protected component
  - fixed branch: the same navigation ran middleware and redirected before mount

Regression coverage includes allowed middleware, aliases and route parameters, auth/request/route context, push, replace, allowed and denied preload, exact redirect and 403 response preservation, exception reporting/rendering, and same-pattern parameter isolation.

### DIFF
--- a/src/Edge/NativeRouter.php
+++ b/src/Edge/NativeRouter.php
@@ -2,6 +2,12 @@
 
 namespace Native\Mobile\Edge;
 
+use Illuminate\Http\Request;
+use Illuminate\Routing\Pipeline;
+use Illuminate\Routing\Route;
+use Illuminate\Support\Facades\Facade;
+use Symfony\Component\HttpFoundation\Response;
+
 class NativeRouter
 {
     /**
@@ -80,9 +86,9 @@ class NativeRouter
      * URI → component class registry.
      * Populated by Route::native() calls.
      *
-     * Each entry: ['class' => string, 'layout' => ?string]
+     * Each entry: ['class' => string, 'layout' => ?string, 'route' => ?Route]
      *
-     * @var array<string, array{class: string, layout: ?string}>
+     * @var array<string, array{class: string, layout: ?string, route: ?Route}>
      */
     protected static array $routes = [];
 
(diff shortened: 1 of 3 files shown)


simonhamp (approved):
Awesome!

simonhamp:
Sorry for the turnaround on this. Superseded by #348

PR #254, opened 2026-08-02
fix(ios): recompute remaining space before placing justified children

shanerbaner82:
## Problem
FlexContainer computes the leftover main-axis space used for justify-content placement **before** Phase 3 sizing finalizes child frames. Justified rows then place children against a stale total — visibly, icons/text in a justified row sit lower than intended (the frame is misplaced; the glyph is correctly centered inside the wrong frame).

## Fix
Recompute `remaining` from the final child sizes immediately before placement.

## Testing
Reproduced with a justified row of icon + text on device; frames land where CSS-equivalent flexbox would put them after the fix. Diagnosed by painting the child frame background to separate frame-placement from glyph-centering.

https://claude.ai/code/session_01KTcfKtsVapKiGfSKwKT69y

### DIFF
--- a/resources/xcode/NativePHP/NativeRender/FlexContainer.swift
+++ b/resources/xcode/NativePHP/NativeRender/FlexContainer.swift
@@ -585,9 +585,20 @@ struct FlexContainer: Layout {
             }
         }
 
-        // Phase 4: Compute justify_content offsets
+        // Phase 4: Compute justify_content offsets.
+        //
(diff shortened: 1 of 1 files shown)


simonhamp:
Can you add before and after screenshots?

shanerbaner82:
> C
justify-end-AFTER-fixed
justify-end-BEFORE-clipped
an you add before and after screenshots?

PR #258, opened 2026-08-02
feat: GatedBridge contract + FakeBridge capability deny-list

shanerbaner82:
## What
`nativephp_can()` consults a new `Contracts\GatedBridge` interface when the active bridge implements it. `FakeBridge` implements it with a `withoutCapability()` deny-list.

## Why
`nativephp_can()` currently answers "available" unconditionally outside a device, so the capability-missing branch of gated app code has never been testable — and Jump sessions claim support for everything regardless of what the connected device registered. This adds the seam:

```php
FakeBridge::enable()->withoutCapability('Camera.GetPhoto');
expect(nativephp_can('Camera.GetPhoto'))->toBeFalse();
```

Default behavior is unchanged everywhere (empty deny-list answers true; bridges that don't implement the contract keep assuming everything is available). Follow-up potential: JumpBridge implementing the contract with the device's actual registered function list.

## Testing
New `GatedBridgeTest`; full suite green.

https://claude.ai/code/session_01KTcfKtsVapKiGfSKwKT69y

### DIFF
--- a/src/Contracts/GatedBridge.php
+++ b/src/Contracts/GatedBridge.php
@@ -1,0 +1,19 @@
+<?php
+
+namespace Native\Mobile\Contracts;
+
+/**
+ * A bridge that can answer capability queries honestly.
+ *
+ * nativephp_can() assumes every method is available (true on device,
+ * and the safe default for Jump / plain FakeBridge tests). A bridge
+ * standing in for the native runtime — a test double stubbing a subset
+ * of methods, a Jump session that knows what the connected device
+ * actually registered — implements this so app code can feature-gate
+ * on what the runtime really provides, without core naming any
+ * concrete bridge class.
+ */
+interface GatedBridge
+{
+    public function can(string $method): bool;
+}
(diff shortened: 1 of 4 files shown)


simonhamp (commented):
I'm not clear why it should be `withoutCapability` feels weird to me to state what something _shouldn't_ have...

I think it would be more idiomatic to act like the real bridge does, by registering features into it, kinda like this:

```php
FakeBridge::withCapabilities([
    'Camera.GetPhoto',
    'Nfc.Scan',
    ...
]);
```

shanerbaner82:
Hey @simonhamp — some context on why this one's worth landing.

`nativephp_can()` currently answers "available" for anything you ask it when running off-device. Two problems with that:

1. **The fallback branch has never been testable.** Any code doing `if (nativephp_can('X')) { ... } else { /* fallback */ }` has an else branch that's unreachable in tests, because every bridge says yes to everything. With this it's one line — `FakeBridge::enable()->withoutCapability('Camera.GetPhoto')` — and the path users actually hit when something's missing is finally under test.

2. **Jump has the same blind spot.** On the dev machine, `nativephp_can()` claims the connected device supports every method, even when the plugin providing it isn't installed. This interface is the missing seam — the natural follow-up is JumpBridge answering from what the device actually registered.

Default behavior changes for nobody: an empty deny-list answers true, and bridges that don't implement the contract keep behaving exactly as before. It's an 18-line interface plus a deny-list on FakeBridge.

CI's green whenever you get a chance to review.

PR #262, opened 2026-08-02
Honour `platforms`, and stop copying SwiftPM manifests and test targets into the app target

ICTtrying:
## The problem

`IOSPluginCompiler::copySwiftFilesRecursively()` copies **every** `.swift` file under a plugin's `resources/ios/` into `ios/NativePHP/Bridge/Plugins/<Namespace>/`, recursively, excluding nothing.

That directory is inside a `PBXFileSystemSynchronizedRootGroup`:

```
95BD5DC32D178E9D00C72704 /* NativePHP */ = {
    isa = PBXFileSystemSynchronizedRootGroup;
    exceptions = (
        /* Exceptions for "NativePHP" folder in "NativePHP" target */,
        /* Exceptions for "NativePHP" folder in "NativePHP-simulator" target */,
    );
```

…and both exception sets list only `Info.plist`. So every file copied under `NativePHP/` becomes a member of the compile sources phase of both app targets automatically, with no per-file opt-out.

Three kinds of file cannot compile in an app target:

| File | Why it fails |
|---|---|
| `Package.swift` | `import PackageDescription` — that module exists only inside SwiftPM's own build |
| `Tests/**/*.swift` | `import XCTest` / `import Testing` and `@testable import` |
| `.build/**/*.swift` | `swift build` leaves generated Swift there |

So a plugin that develops its platform-neutral Swift as a Swift Package — the normal way to unit-test Swift logic outside Xcode, and on a machine that isn't a Mac — breaks the iOS build of **every application that installs it**.

The second half is that the decision to copy at all comes from `Plugin::hasIosCode()`, which is true if any `.swift` file exists anywhere under `resources/ios/`. It never consults the manifest's `platforms`. `native:plugin:create` has emitted that key for a long time (`PluginCreateCommand.php:341`) and the generated plugin test asserts its values (`:1016`), but nothing in core reads it — `PluginManifest` does not even keep it. An Android-only plugin therefore still contributes sources, bridge registrations and `Info.plist` keys to an iOS build.

## Reproduction

Runs on Linux; the damage is done in PHP long before `xcodebuild`. Full script in the collapsed block below — it creates a clean Laravel app, installs a plugin whose `resources/ios/` holds a Swift Package, scaffolds the Xcode project the way `InstallsIos::createXcodeProject()` does, and calls the same `IOSPluginCompiler::compile()` that `BuildIosAppCommand` calls.

The plugin's manifest says:

```json
{ "namespace": "Hello", "platforms": ["android"] }
```

**Before:**

```
nativephp/ios/NativePHP/Bridge/Plugins/Hello/HelloCore/.build/debug/HelloCore.build.swift
nativephp/ios/NativePHP/Bridge/Plugins/Hello/HelloCore/Package.swift
nativephp/ios/NativePHP/Bridge/Plugins/Hello/HelloCore/Sources/HelloCore/Hello.swift
nativephp/ios/NativePHP/Bridge/Plugins/Hello/HelloCore/Tests/HelloCoreTests/HelloTests.swift
nativephp/ios/NativePHP/Bridge/Plugins/Hello/HelloFunctions.swift
```

Five files, from a plugin that declares no iOS support at all. `Package.swift` and `HelloTests.swift` both fail to compile.

**After:** nothing is copied.

**After, with the same plugin declaring `"platforms": ["android", "ios"]`:**

```
nativephp/ios/NativePHP/Bridge/Plugins/Hello/HelloFunctions.swift
```

The bridge file — which is the plugin's actual iOS surface — and nothing else.

## The fix

**`platforms` is read, validated and honoured**, by both compilers. Absent, null or empty means both platforms, so no existing plugin changes behaviour. An unknown value throws rather than being dropped: silently ignoring `"IOS"` would turn a typo into "this plugin has no iOS code", which is the failure the key exists to prevent.

**`SwiftSourceFilter`** holds the exclusions in one place — the copier, `hasIosCode()` and `getIosSourceFiles()` all use it, so they cannot disagree about what a plugin's iOS sources are. A nested directory containing a `Package.swift` is skipped as a subtree, because Xcode consumes embedded packages as package dependencies rather than as loose sources; copying them file by file duplicates every symbol.

**`ios.sources`** is a new optional allow-list, relative to the plugin's iOS source path, naming files or directories. When present it is the whole list. It is the escape hatch for a layout the automatic exclusions get wrong, and it makes a plugin's iOS surface explicit and reviewable.

Hooks are deliberately left alone: a hook is the plugin's own code and gets to decide for itself whether it has anything to do on a given platform.

## Tests

13 new tests. All of them fail on `main` and pass here:

- `IOSCompilerTest`: an embedded Swift package is not copied; a `Tests/` directory is not copied; a plugin without declared iOS support contributes no sources, no registrations and no `Info.plist` keys; a plugin with no `platforms` key still gets its sources copied (the regression guard); `ios.sources` copies exactly what it names.
- `PluginTest`: `platforms` defaults to both, is normalised for case and duplicates, rejects unknown values, and gates `hasIosCode()`/`hasAndroidCode()` and the source-file listings in both directions.

`vendor/bin/pest`, `vendor/bin/pint --test` and `vendor/bin/phpstan analyse` are all clean. (Two pre-existing failures in `ReleaseBuildBundleTest` are a missing local `ZipArchive` extension and are identical on `main`.)

## Not in this PR

- **The Android analogue of the exclusions.** `platforms` now gates the Android compiler too, but `AndroidPluginCompiler` maps Kotlin files by their `package` declaration rather than by path, so a path allow-list there would mean something different. A plugin shipping a Gradle module under `resources/android/` with a `src/test/` tree has the same class of problem and deserves its own change.
- **A `native:plugin:validate` warning** when `resources/ios/` holds Swift that `platforms` excludes. Useful, but it is a new diagnostic rather than a fix.

<details>
<summary>Reproduction script</summary>

```bash
#!/usr/bin/env bash
#
# Reproduce: a plugin that ships a Swift Package under resources/ios/ has that
# package's Package.swift and test target copied into the host app's iOS target.
#
# Runs on Linux. It does not need Xcode: the failure happens in PHP, in
# IOSPluginCompiler, before xcodebuild is ever invoked.
#
# Usage: ./reproduce.sh /path/to/empty/dir
set -euo pipefail

WORK="${1:?usage: reproduce.sh <empty-dir>}"
mkdir -p "$WORK"
cd "$WORK"

# ---------------------------------------------------------------- 1. clean app
cd app

# ------------------------------------------------- 2. a plugin with a SwiftPM package
#
# Shape is the one `native:plugin:create` generates (flat resources/ios/) plus a
# Swift Package alongside it — the layout you use when platform-neutral logic is
# developed and unit-tested outside Xcode.
PLUGIN=packages/acme-hello
mkdir -p "$PLUGIN/src" \
         "$PLUGIN/resources/ios/HelloCore/Sources/HelloCore" \
         "$PLUGIN/resources/ios/HelloCore/Tests/HelloCoreTests" \
         "$PLUGIN/resources/ios/HelloCore/.build/debug"

cat > "$PLUGIN/composer.json" <<'JSON'
{
    "name": "acme/hello",
    "version": "1.0.0",
    "type": "nativephp-plugin",
    "autoload": { "psr-4": { "Acme\\Hello\\": "src/" } },
    "extra": { "laravel": { "providers": ["Acme\\Hello\\HelloServiceProvider"] } }
}
JSON

# platforms says android only. The compiler never reads it.
cat > "$PLUGIN/nativephp.json" <<'JSON'
{
    "name": "acme/hello",
    "namespace": "Hello",
    "platforms": ["android"],
    "bridge_functions": [
        { "name": "Hello.Ping", "android": "com.acme.hello.Ping" }
    ]
}
JSON

cat > "$PLUGIN/src/HelloServiceProvider.php" <<'PHP'
<?php

namespace Acme\Hello;

use Illuminate\Support\ServiceProvider;

class HelloServiceProvider extends ServiceProvider {}
PHP

cat > "$PLUGIN/resources/ios/HelloFunctions.swift" <<'SWIFT'
import Foundation

// A normal bridge file. This one *should* be copied — on iOS.
enum HelloFunctions {}
SWIFT

cat > "$PLUGIN/resources/ios/HelloCore/Package.swift" <<'SWIFT'
// swift-tools-version: 5.9
import PackageDescription

let package = Package(
    name: "HelloCore",
    products: [.library(name: "HelloCore", targets: ["HelloCore"])],
    targets: [
        .target(name: "HelloCore"),
        .testTarget(name: "HelloCoreTests", dependencies: ["HelloCore"]),
    ]
)
SWIFT

cat > "$PLUGIN/resources/ios/HelloCore/Sources/HelloCore/Hello.swift" <<'SWIFT'
import Foundation

public enum Hello {
    public static func greet() -> String { "hi" }
}
SWIFT

cat > "$PLUGIN/resources/ios/HelloCore/Tests/HelloCoreTests/HelloTests.swift" <<'SWIFT'
import XCTest
@testable import HelloCore

final class HelloTests: XCTestCase {
    func testGreet() { XCTAssertEqual(Hello.greet(), "hi") }
}
SWIFT

# SwiftPM build residue. `swift build` leaves generated .swift files here.
cat > "$PLUGIN/resources/ios/HelloCore/.build/debug/HelloCore.build.swift" <<'SWIFT'
// generated build artifact
let generated = true
SWIFT

php -r '
$f = "composer.json";
$j = json_decode(file_get_contents($f), true);
$j["repositories"] = [["type" => "path", "url" => "packages/acme-hello", "options" => ["symlink" => false]]];
$j["require"]["acme/hello"] = "*";
file_put_contents($f, json_encode($j, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES) . "\n");
'
composer update acme/hello --no-interaction --quiet

# Plugins are opt-in: publish the provider, then register.
php artisan vendor:publish --tag=nativephp-plugins-provider --quiet
php artisan native:plugin:register acme/hello --quiet

# ------------------------------------------------- 3. scaffold the Xcode project
# Exactly what InstallsIos::createXcodeProject() does, minus the PHP binaries
# (which need no part in this).
php -r '
mkdir("nativephp/ios", 0755, true);
$src = "vendor/nativephp/mobile/resources/xcode";
foreach ($it as $item) {
    $dest = "nativephp/ios/" . substr($item->getPathname(), strlen($src) + 1);
}
'

# ------------------------------------------------- 4. run the iOS plugin compile
# This is the same call BuildIosAppCommand makes (line 1213/1226). It is the
# whole reproduction: no Mac needed, the damage is done in PHP.
php -r '
require "vendor/autoload.php";
$app = require "bootstrap/app.php";
$app->make(Illuminate\Contracts\Console\Kernel::class)->bootstrap();
$app->make(Native\Mobile\Plugins\Compilers\IOSPluginCompiler::class)->compile();
'

echo
echo "=== what landed in the app target ==="
find nativephp/ios/NativePHP/Bridge/Plugins -type f | sort
```

</details>

### DIFF
--- a/src/Plugins/Compilers/AndroidPluginCompiler.php
+++ b/src/Plugins/Compilers/AndroidPluginCompiler.php
@@ -134,13 +134,11 @@ class AndroidPluginCompiler
             throw new PluginConflictException($conflicts);
         }
 
-        $allPlugins = $this->registry->all();
-        $hookRunner = $this->getHookRunner();
-
         // A plugin that declares `platforms: ["ios"]` contributes nothing to an
         // Android build. Hooks still run for every plugin — a hook is the
         // plugin's own code and gets to decide for itself.
-        $allPlugins = $allPlugins->filter(fn (Plugin $p) => $p->supportsPlatform('android'));
+        $allPlugins = $this->registry->all()->filter(fn (Plugin $p) => $p->supportsPlatform('android'));
+        $hookRunner = $this->getHookRunner();
 
         // Run pre-compile hooks
         $hookRunner->runPreCompileHooks();
--- a/src/Plugins/Compilers/AndroidPluginCompiler.php
+++ b/src/Plugins/Compilers/AndroidPluginCompiler.php
@@ -137,6 +137,11 @@ class AndroidPluginCompiler
         $allPlugins = $this->registry->all();
         $hookRunner = $this->getHookRunner();
 
+        // A plugin that declares `platforms: ["ios"]` contributes nothing to an
+        // Android build. Hooks still run for every plugin — a hook is the
+        // plugin's own code and gets to decide for itself.
+        $allPlugins = $allPlugins->filter(fn (Plugin $p) => $p->supportsPlatform('android'));
+
         // Run pre-compile hooks
         $hookRunner->runPreCompileHooks();
 
--- a/src/Plugins/Compilers/IOSPluginCompiler.php
+++ b/src/Plugins/Compilers/IOSPluginCompiler.php
@@ -8,6 +8,7 @@ use Native\Mobile\Exceptions\PluginConflictException;
 use Native\Mobile\Plugins\Plugin;
 use Native\Mobile\Plugins\PluginHookRunner;
 use Native\Mobile\Plugins\PluginRegistry;
+use Native\Mobile\Plugins\SwiftSourceFilter;
 use Native\Mobile\Support\Stub;
 
 class IOSPluginCompiler
@@ -96,6 +97,12 @@ class IOSPluginCompiler
         $allPlugins = $this->registry->all();
         $hookRunner = $this->getHookRunner();
 
+        // A plugin that declares `platforms: ["android"]` contributes nothing
+        // to an iOS build: no sources, no registrations, no Info.plist keys.
+        // Hooks still run for every plugin — a hook is the plugin's own code
+        // and gets to decide for itself.
+        $iosPlugins = $allPlugins->filter(fn (Plugin $p) => $p->supportsPlatform('ios'));
+
         // Run pre-compile hooks
         $hookRunner->runPreCompileHooks();
 
@@ -108,10 +115,10 @@ class IOSPluginCompiler
         $this->files->ensureDirectoryExists($this->generatedPath);
 
         // Get plugins with iOS code (for copying files)
-        $pluginsWithCode = $allPlugins->filter(fn (Plugin $p) => $p->hasIosCode());
+        $pluginsWithCode = $iosPlugins->filter(fn (Plugin $p) => $p->hasIosCode());
 
         // Get plugins with iOS bridge functions (for registration)
-        $pluginsWithFunctions = $allPlugins->filter(function (Plugin $p) {
+        $pluginsWithFunctions = $iosPlugins->filter(function (Plugin $p) {
             $functions = $p->getBridgeFunctions();
             foreach ($functions as $function) {
                 if (! empty($function['ios'])) {
@@ -123,12 +130,12 @@ class IOSPluginCompiler
         });
 
         // Get plugins with iOS info_plist entries or dependencies
-        $pluginsWithIosData = $allPlugins->filter(function (Plugin $p) {
+        $pluginsWithIosData = $iosPlugins->filter(function (Plugin $p) {
             return ! empty($p->getIosInfoPlist()) || ! empty($p->getIosDependencies());
         });
 
         // Check for plugins with iOS UI component renderers
-        $pluginsWithRenderers = $allPlugins->filter(function (Plugin $p) {
+        $pluginsWithRenderers = $iosPlugins->filter(function (Plugin $p) {
             foreach ($p->getComponents() as $component) {
                 if (! empty($component['ios_renderer'])) {
                     return true;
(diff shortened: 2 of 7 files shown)


src/Plugins/Compilers/AndroidPluginCompiler.php
```diff
@@ -137,6 +137,11 @@ public function compile(): void
         $allPlugins = $this->registry->all();
         $hookRunner = $this->getHookRunner();
 
+        // A plugin that declares `platforms: ["ios"]` contributes nothing to an
+        // Android build. Hooks still run for every plugin — a hook is the
+        // plugin's own code and gets to decide for itself.
+        $allPlugins = $allPlugins->filter(fn (Plugin $p) => $p->supportsPlatform('android'));
```
simonhamp: As we're overwriting `$allPlugins`, shouldn't we just apply this directly to line 137?

src/Plugins/Compilers/IOSPluginCompiler.php
```diff
@@ -96,6 +97,12 @@ public function compile(): void
         $allPlugins = $this->registry->all();
         $hookRunner = $this->getHookRunner();
 
+        // A plugin that declares `platforms: ["android"]` contributes nothing
+        // to an iOS build: no sources, no registrations, no Info.plist keys.
+        // Hooks still run for every plugin — a hook is the plugin's own code
+        // and gets to decide for itself.
+        $iosPlugins = $allPlugins->filter(fn (Plugin $p) => $p->supportsPlatform('ios'));
```
simonhamp: Weirdly inconsistent naming/renaming here. In `AndroidPluginCompiler` we override `$allPlugins` - here we copy to `$iosPlugins`

simonhamp (commented):
Looks really reasonable. Some low-risk style questions. Let's tighten this up and we can get it shipped out

src/Plugins/Compilers/AndroidPluginCompiler.php
ICTtrying: Done — folded into the initial assignment, so it now reads:

```php
$allPlugins = $this->registry->all()->filter(fn (Plugin $p) => $p->supportsPlatform('android'));
```

Also brought iOS in line with this shape rather than the other way round (see reply on that thread) — both compilers now filter into `$allPlugins` the same way. Pushed in 0ca3f38; the 13 new tests and the rest of the suite (811 tests total) are still green, and pint/phpstan are clean.

src/Plugins/Compilers/IOSPluginCompiler.php
ICTtrying: Matched Android's shape here rather than the reverse: iOS now overwrites `$allPlugins` in place (filter folded into the initial `->all()` call) instead of copying into `$iosPlugins`:

```php
$allPlugins = $this->registry->all()->filter(fn (Plugin $p) => $p->supportsPlatform('ios'));
```

That meant reverting the 12 call sites below back to `$allPlugins` too. Went this way rather than renaming Android's variable throughout, since `$allPlugins` is what the rest of Android's `compile()` — and this file, pre-PR — already used everywhere; overwriting in place keeps that and is the smaller diff. Pushed in 0ca3f38, 13 new tests + full suite (811 tests) still green.

simonhamp (approved):
Nice

PR #265, opened 2026-08-02
Accept format v4 — event channel is a queue, and pin the binaries release

simonhamp:
App-side half of the format v4 bump. The extension-side change — NativePHP/php-bin-mobile#20, "Make the native→PHP event channel a queue" — turns the single-slot native→PHP event channel into a FIFO queue. This PR updates the iOS and Android readers to accept it, and pins which release of the PHP binaries we install.

> [!WARNING]
> **Do not merge or release this until php-bin-mobile#20 is released.**
>
> `4.0.0.json` does not exist on `bin.nativephp.com` yet. Until that release is published, this branch pins a manifest that 404s and every `native:install` fails at the fetch. php-bin-mobile#20 is still a draft.

## Why these two changes are one change

They look unrelated — a version constant and a URL — but the second exists because of the first.

Until now the installer fetched `{branch}/versions.json`, a floating name: whatever binaries were published most recently. That was survivable only while every published build spoke the same wire format. **This is the first time that stops being true.** Once v3 and v4 binaries both exist, a floating manifest hands an install whichever was published last, with no relationship to the `expectedFormatVersion` compiled into the Swift and Kotlin readers sitting next to it.

The failure that produces is the quiet kind. The readers compare the format number on the first publish and refuse to render on a mismatch — so the app launches, runs, and simply never draws anything. No crash, no error, nothing pointing at the binaries.

So the version handshake moving to 4 is what forces the pin. They ship together or the pin is pointless.

## The format change

**Version handshake → 4** — `expectedFormatVersion` (`NativeElementBridge.swift`), `EXPECTED_FORMAT_VERSION` (`NativeElementBridge.kt`).

Nothing changes for either reader: the per-frame wire format is identical to v3, and the flat/prop render-direction buffers are untouched. The version moved because the *region's event fields changed meaning*, so a stale `libphp.a` fails loud rather than being quietly wrong.

The motivating bug: before v4, a second event posted before PHP drained the first **silently overwrote it**. The async-task lane in NativePHP/php-bin-mobile#228 hits this hard — five posting threads against one slot.

**`nphp_element_post_event` is now `int`, not `void`** — `PHP.h`, `bridge_jni.cpp`.

Returns `1` if the event was queued, `0` if dropped (no region, or the queue is over its backlog cap because PHP stopped draining). Both declarations had to move to match the definition in `nphp_element.c`. Swift's `nativeElementWriteEvent` explicitly discards the result — a UI event has nothing useful to do about a drop, and the extension already logs it with the `callback_id` it would have woken — with a note that callers who *do* care (async-task completions) should check rather than copy that.

**Region event mirrors marked dead** — `NativeElementBridge.swift` offsets, `NphpElementRegion` in `bridge_jni.cpp`.

`event_size` / `event_count` / `event_buffer` are kept only so the surrounding offsets stay correct. Under a queue of separately allocated frames, `event_buffer` is never written, `event_size` is the head frame's size rather than "the" event's, and `event_count` is a depth rather than a 0/1 flag. Writing there is silently ignored; reading gives garbage. The JNI struct comment also now spells out that it is only a *prefix* of the real `nphp_element_region_t` — appending to the real struct is safe, inserting above `current_tree` is not.

## The pin

**New `src/Support/PhpBinaries.php`** — holds `VERSION = '4.0.0'`, matching `NATIVEPHP_VERSION` in php-bin-mobile's `config/versions.sh`, and builds the manifest URL from it. Install now fetches `{branch}/4.0.0.json` instead of `{branch}/versions.json`.

`composer.lock` already pins this package, so the binaries are transitively pinned too — nothing to record in `nativephp.lock`. The class docblock spells out the bump rule: whenever you want the installed binaries to change, and *mandatorily* whenever the new release moves `NPHP_FORMAT_VERSION`, in which case the two reader constants have to move in the same release.

**Sharper failures**, because pinning introduces a new way to fail and the old messages misdirect:

- A 404 on the manifest now means something specific — the manifest is named for the release this package pins, so a missing one means that release was withdrawn, not that the CDN is down. It says so and points at `composer update nativephp/mobile`.
- When the manifest is missing entirely, the iOS and Android installers return quietly instead of adding `PHP {version} binaries not available`. The fetch already explained itself, and that message blames the PHP version, which isn't the problem.
- When the manifest is present but lacks the requested PHP version, name the release it was missing from.

## Testing

`tests/Unit/Support/PhpBinariesTest.php` — 4 tests covering the URL shape, branch handling, that the name is *not* floating, and that `VERSION` is semver. All passing.

The format side has no behaviour change beyond the version constant — same event path, framing, and buffers. It needs a build against the v4 `libphp.a` to confirm the handshake passes on both platforms, which is blocked on the release above.

### DIFF
--- a/docs/event-channel-pointer-refactor.md
+++ b/docs/event-channel-pointer-refactor.md
@@ -193,6 +193,13 @@ Widened (event channel only — render-direction flat/prop buffers untouched):
 4. **Version handshake** — `NPHP_FORMAT_VERSION` (nphp_element.h),
    `expectedFormatVersion` (Swift), `EXPECTED_FORMAT_VERSION` (Kotlin) all → 3.
 
+> These three are on **4** now, not 3. v4 turned the event channel into a FIFO
+> queue: before it, a second post landing before PHP drained the first
+> overwrote it silently, which the async-task lane in #228 hit hard (five
+> posting threads). The per-frame wire format is identical to v3, so nothing
+> in the widening described above changed — only the region's event fields and
+> the version they're guarded by.
+
 `tabChange` keeps its u16 (it's an index, not a length).
 
 Render-direction caps (`npui_write_str` u16, node `prop_size`/`raw_children` u16)
--- a/resources/androidstudio/app/src/main/cpp/bridge_jni.cpp
+++ b/resources/androidstudio/app/src/main/cpp/bridge_jni.cpp
@@ -42,7 +42,10 @@ static void element_write_event(JNIEnv*, jclass, jint, jint, jint, jbyteArray);
 extern "C" uint32_t nphp_get_format_version(void);
 extern "C" uint32_t nphp_get_runtime_flags(void);
 extern "C" void     nphp_set_runtime_flags(uint32_t flags);
-extern "C" void     nphp_element_post_event(int type, int callback_id, int node_id,
+// Returns 1 if the event was queued, 0 if dropped (no region, or the queue is
+// over its backlog cap because PHP stopped draining). Was void before format
+// v4 — the declaration has to match the definition in nphp_element.c.
+extern "C" int      nphp_element_post_event(int type, int callback_id, int node_id,
                                             const uint8_t *data, uint32_t data_len);
 
 // Phase 3 — active-buffer accessors. These do the acquire-load on the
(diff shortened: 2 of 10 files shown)


simonhamp:
I have manually tested that this runs and installs successfully on both iOS and Android simulators with the new binaries

Issue #251, opened 2026-08-01
v4: Android build fails with "Plugin with id 'com.google.gms.google-services' not found" when google-services.json is present


aotearoait:
### What happened?

On v4.0.1, an Android build fails immediately for any app that ships a `google-services.json` (i.e. any app using `nativephp/mobile-firebase`):

```
FAILURE: Build failed with an exception.

* Where:
Build file 'nativephp/android/app/build.gradle.kts' line: 9

* What went wrong:
Plugin with id 'com.google.gms.google-services' not found.
```

The generated `app/build.gradle.kts` applies the plugin conditionally (this comes straight from the shipped template, `resources/androidstudio/app/build.gradle.kts`):

```kotlin
val googleServicesJson = file("google-services.json")
if (googleServicesJson.exists()) {
    apply(plugin = "com.google.gms.google-services")
}
```

…but the plugin is never put on the classpath. In the shipped template:

- `resources/androidstudio/build.gradle.kts` declares only `android.application`, `kotlin.android` and `kotlin.compose`.
- `resources/androidstudio/gradle/libs.versions.toml` has no `google-services` entry under `[plugins]`.
- `resources/androidstudio/settings.gradle.kts` has no `pluginManagement` entry resolving it.

So `apply(plugin = "com.google.gms.google-services")` has nothing to resolve against and Gradle fails at configuration time.

### How to reproduce

1. `composer require nativephp/mobile:~4.0.0 nativephp/mobile-firebase`
2. `php artisan native:plugin:register nativephp/mobile-firebase`
3. Drop a real `google-services.json` in the project root
4. `php artisan native:install --force`
5. Build for Android

Configuration fails before any compilation. Removing `google-services.json` builds fine, but then Firebase push is not wired up, so that is not a workaround.

### Expected

The Android project configures and builds with Firebase push enabled.

### Workaround

Declare the plugin in the generated top-level `nativephp/android/build.gradle.kts`:

```kotlin
plugins {
    alias(libs.plugins.android.application) apply false
    alias(libs.plugins.kotlin.android) apply false
    alias(libs.plugins.kotlin.compose) apply false
    id("com.google.gms.google-services") version "4.4.2" apply false
}
```

Because `nativephp/` is generated, this has to be re-applied after every `native:install --force`.

### Suggested fix

[...]

PR #274, opened 2026-08-03
Declare the Google Services Gradle plugin the app module already applies

ICTtrying:
## The problem

`resources/androidstudio/app/build.gradle.kts` applies the Google Services Gradle plugin whenever the project has a `google-services.json`:

```kotlin
val googleServicesJson = file("google-services.json")
if (googleServicesJson.exists()) {
    apply(plugin = "com.google.gms.google-services")
}
```

and `com.google.gms:google-services` appears on **no classpath anywhere in core** — not the root `build.gradle.kts`, not `settings.gradle.kts`, not `gradle/libs.versions.toml`. The conditional reads exactly right and fires into nothing:

```
* Where: nativephp/android/app/build.gradle.kts line: 9
* What went wrong: Plugin with id 'com.google.gms.google-services' not found.
```

Every Firebase feature requires that file, so the conditional cannot fire without breaking the build. `RunsAndroid::updateFirebaseConfiguration()` copies the file in on every build, so this is core carrying its own feature up to the last step and then stopping. `PushNotifications` is core's own facade, which makes this core's own push story failing at the first build.

## Reproduction

`docs/upstream/reproduce-google-services.sh <empty-dir>` — no device, no Android SDK, no Gradle. It creates a clean Laravel app, requires `nativephp/mobile:^4.0`, scaffolds the Android project the way `InstallsAndroid::createAndroidStudioProject()` does, adds a `google-services.json`, and reads the generated files:

```
=== what the app module does with it ===
7:val googleServicesJson = file("google-services.json")
8-if (googleServicesJson.exists()) {
9-    apply(plugin = "com.google.gms.google-services")
10-}

=== where the plugin is declared ===
nowhere: not the root build file, not settings.gradle.kts, not the version catalog.
```

<details><summary>The script</summary>

```bash
#!/usr/bin/env bash
set -euo pipefail

WORK="${1:?usage: reproduce-google-services.sh <empty-dir>}"
mkdir -p "$WORK"; cd "$WORK"

cd app

# What InstallsAndroid::createAndroidStudioProject() does, minus the PHP binaries.
php -r '
mkdir("nativephp/android", 0755, true);
$src = "vendor/nativephp/mobile/resources/androidstudio";
foreach ($it as $item) {
    $dest = "nativephp/android/" . substr($item->getPathname(), strlen($src) + 1);
}
'

echo '{ "project_info": { "project_id": "reproduction" } }' > google-services.json
cp google-services.json nativephp/android/app/google-services.json

grep -n -A3 'val googleServicesJson' nativephp/android/app/build.gradle.kts
grep -rn 'com\.google\.gms[.:]google-services' \
    nativephp/android/build.gradle.kts \
    nativephp/android/settings.gradle.kts \
    nativephp/android/gradle/libs.versions.toml || echo "declared nowhere"
```

</details>

### And with Gradle, on a real project

The script stops at the state because an Android SDK is a big ask of a reviewer. The build itself was run on a real app (Galaxy S23, `nativephp/mobile 4.0.1`, the app's own `google-services.json` from the Firebase CLI):

| Root `build.gradle.kts` | Result |
|---|---|
| The shipped template, untouched | `BUILD FAILED in 2s` — `Plugin with id 'com.google.gms.google-services' not found.` at `app/build.gradle.kts` line 9 |
| The same tree, plus the declaration this PR writes | `BUILD SUCCESSFUL in 49s`, APK installed |

Nothing else differed between the two runs.

## The fix

One method, called from the one that already copies `google-services.json` in.

**It writes the generated root build file, not the template.** That is the part worth reviewing rather than the regex. `resources/androidstudio/` is copied by `native:install` and nothing rewrites it afterwards, so a template-only fix repairs projects created *after* it ships and leaves every existing project failing exactly as before. Patching core's template first and watching the second build fail identically to the first is how this was found.

**It is conditional both ways.** The declaration goes in when the app module has a `google-services.json` and comes back out when it does not — `apply false` still resolves the plugin marker, so an app that has nothing to do with Firebase should not be made to fetch one. The block is marker-delimited and rebuilt from the current state on every build, the same shape `injectGradlePlugins` and the ProGuard rules block already use, so it is idempotent and a build file is not a different file after every build.

**An existing declaration wins**, in either shape:

- a plugin's own `android.gradle_plugins` entry, which `AndroidPluginCompiler::injectGradlePlugins()` writes into the same `plugins {}` block;
- a hand-written `classpath("com.google.gms:google-services:…")` in a `buildscript` block — the pre-plugins-DSL shape, and exactly what people add when they hit this.

Declaring the same plugin id twice is a Gradle error, so both are left alone. The plugin compiler already skips ids declared outside its own markers, and a test pins that it sees this one.

**And when it cannot, it says so.** A root build file with no `plugins {}` block gets a warning naming the line to add by hand, rather than a silent no-op — which is the failure mode this whole PR is about, one layer further out.

## Tests

`tests/Unit/GoogleServicesClasspathTest.php`, 14 tests. They run the real `RunsAndroid::updateFirebaseConfiguration()` against a project on disk — worth noting, because `ConfigurationUpdatesTest` reimplements that method locally, so core's own copy step had no test exercising it either.

| | On `main` | On this branch |
|---|---|---|
| 8 tests | **fail** | pass |
| 6 regression guards | pass | pass |

The eight that fail are the behaviour being added: the declaration appears, it is inside the `plugins {}` block, an existing project is repaired, it is removed again when the config file goes, five builds leave one declaration and a byte-identical file, the plugin compiler does not add a second, the no-`plugins`-block case warns, and `nativephp/resources/google-services.json` counts as a source too.

The six that pass on both are the guards: the copy still happens and still copies the right bytes, a project without Firebase gets nothing added and no warning, an existing plugin declaration is left alone, a hand-written buildscript classpath is left alone, and a project with no Android folder is a no-op rather than a crash.

| Check | Result |
|---|---|
| `vendor/bin/pest`, before | 816 pass, 0 fail |
| `vendor/bin/pest`, after | 830 pass, 0 fail |
| `vendor/bin/pint --test` | passed |
| `vendor/bin/phpstan analyse` | no errors |

## Not in this PR

- **The plugin version is a constant** (`4.4.3`, the version core's own test fixtures use), not a config key. A published `config/nativephp.php` does not gain keys on upgrade, so an escape hatch here needs a `config(…, $default)` read and a documented key, and that is a small feature rather than this fix.
- **`PushNotification.*` has no Android bridge implementation.** `PushNotifications::getToken()` calls `nativephp_call('PushNotification.GetToken')`; the Android bridge registers `Camera.*`, `Device.*`, `Dialog.*`, `File.*`, `Location.*`, `Perf.*`, `System.*` and `UI.*`, and nothing push. With this PR a Firebase app builds and its token still comes back null on Android. That is a separate, larger gap and it deserves its own change.
- **iOS.** `GoogleService-Info.plist` has no equivalent conditional to be broken by, and nothing here touches the iOS build.
- **Reformatting the template.** `resources/androidstudio/build.gradle.kts` is untouched on purpose, so that a project which never adds a `google-services.json` resolves exactly the plugins it does today.

### DIFF
--- a/src/Concerns/RunsAndroid.php
+++ b/src/Concerns/RunsAndroid.php
@@ -21,6 +21,21 @@ trait RunsAndroid
 {
     use PreparesBuild, WatchesAndroid;
 
+    /**
+     * The Gradle plugin that processes google-services.json into Android
+     * resources. Applied by the app module whenever that file is present.
+     */
+    private const GOOGLE_SERVICES_PLUGIN_ID = 'com.google.gms.google-services';
+
+    private const GOOGLE_SERVICES_PLUGIN_VERSION = '4.4.3';
+
+    /**
+     * The same plugin as a buildscript classpath coordinate — the shape the
+     * Firebase documentation used before the plugins DSL, and the one people
+     * reach for when they patch this in by hand.
+     */
+    private const GOOGLE_SERVICES_ARTIFACT = 'com.google.gms:google-services';
+
     protected string $androidLogPath = 'nativephp'.DIRECTORY_SEPARATOR.'android-build.log';
 
     /**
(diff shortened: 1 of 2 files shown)


jaeheeship:
Confirmed this issue on a real project using:

nativephp/mobile 4.1.0
nativephp/mobile-firebase 1.1.0
Android / Gradle 8.14.5

Once google-services.json was present, php artisan native:run android failed at app/build.gradle.kts:9 with:

Plugin with id 'com.google.gms.google-services' not found.

Adding the following declaration to the generated root nativephp/android/build.gradle.kts resolved plugin resolution and allowed Gradle project configuration to complete successfully:

id("com.google.gms.google-services") version "4.4.3" apply false

I also confirmed that mobile-firebase 1.1.0 does not declare this through android.gradle_plugins, so AndroidPluginCompiler does not inject it.

This PR’s approach is preferable to the manual workaround because it repairs existing generated Android projects and avoids duplicate declarations. Thanks for fixing this.

simonhamp (approved):
Looks good. Thanks for doing this

shanerbaner82:
Thanks for the thorough investigation, reproduction, and test coverage here. The underlying issue was valid for mobile-firebase 1.1.0, but the Google Services Gradle declaration has since shipped in mobile-firebase 1.1.1 through its android.gradle_plugins manifest.

We want platform-specific build requirements, including Gradle plugin IDs, versions, and application behavior, to remain the responsibility of the plugin that needs them. Other plugins may require similar or entirely different mechanisms, while core should provide the generic compiler and extension points needed to support them.

Closing this as superseded by the package-side fix. If we discover that the plugin manifest cannot express a required behavior, we should address that separately as a generic plugin-system capability rather than adding a Firebase-specific path to core.

simonhamp:
An interim fix is in place via #410, released as v4.3.1. Please upgrade `nativephp/mobile` @ICTtrying @jaeheeship

simonhamp:
This may be related to #251 also.

PR #277, opened 2026-08-03
feat(ios): implement UI.SetBackground for parity with Android

shanerbaner82:
## Problem

`UI.SetBackground` is implemented on Android only (`UIFunctions.kt` paints the decor view). On iOS the call silently no-ops — and separately, iOS has no way to control what renders behind screen content. Dark-themed apps get a **white band in the bottom safe-area inset on every NavigationStack-hosted screen**: `NavigationStack` draws its own `systemBackground` container and SwiftUI exposes no override hook for it. The only app-side workaround today is moving dark screens out of stack layouts entirely.

## Fix

Adds the iOS handler with the same contract as Android (`{color: "#0F172A"}` → `{success: true}`, accepting `#RGB` / `#RRGGBB` / `#AARRGGBB` per `Color.parseColor`). The color lands in a new `WindowBackgroundState` observable read by three surfaces, all falling back to `systemBackground` when unset — zero behavior change for apps that never call it:

1. **`ContentView`** — the per-screen base layer and the boot placeholder, so the color shows during screen transitions (no more white flash between dark screens) and before first content.
2. **`NativeRootStackRenderer`** — each stack screen backgrounds itself with the color, extended through the safe areas (no-op-when-unset `ViewModifier`), fixing the white bottom band.
3. **UIKit windows** — so overscroll and rotation gaps match.

(`UI.SetTransition`, Android's other UI.* function, intentionally has no iOS twin — transitions ride the tree publish path there.)

## Testing

- Template compiles clean: 284 Swift files, zero errors (simulator SDK; the linker step needs the app-embedded workspace, unrelated to this change).
- Verified in the demo app that dark screens calling `UI.SetBackground` in `mount()` previously showed the white bottom band on device.

### DIFF
--- a/resources/androidstudio/app/src/main/java/com/nativephp/mobile/bridge/functions/UIFunctions.kt
+++ b/resources/androidstudio/app/src/main/java/com/nativephp/mobile/bridge/functions/UIFunctions.kt
@@ -1,6 +1,7 @@
 package com.nativephp.mobile.bridge.functions
 
 import android.graphics.Color
+import android.graphics.drawable.Drawable
 import android.os.Handler
 import android.os.Looper
 import androidx.fragment.app.FragmentActivity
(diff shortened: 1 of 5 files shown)


shanerbaner82:
Follow-up commit: testing in the demo app surfaced that the call is app-global sticky state — once a dark screen sets a color, every screen visited afterwards inherits it (this leak existed on Android all along, it just became visible once iOS honored the call). A `null`/missing/empty color now **clears** the override on both platforms: iOS resets to `systemBackground`; Android captures the decor view's pre-override drawable on first set and restores it on clear (previously there was no way back to the theme default). Screens that set a background should clear it in `unmount()`.

PR #280, opened 2026-08-03
Fix - Expose tree publication revisions on Android

### LINKED ISSUE #279: Android plugin renderers cannot observe identical native tree publications

Interactive native controls update immediately for a responsive feel. If the server rejects that change and republishes the previous, structurally identical tree, Android currently ignores the publication and the control remains visually out of sync. Every server publication must be observable - even when its value is unchanged -so optimistic controls can reliably reconcile with authoritative state.

On Android, a successfully mounted native UI tree is exposed through a Compose `MutableState<NativeUITree?>`. Compose suppresses state writes when the new tree compares equal to the current tree. As a result, plugin renderers cannot observe a valid server publication when its tree is structurally identical to the preceding tree.

This creates a platform contract mismatch. A publication is an event that can carry authoritative reconciliation meaning even when its payload is unchanged. Android currently exposes only whether the payload changed.

## Affected code
- `NativeUIBridge.currentTree` is created with `mutableStateOf<NativeUITree?>(null)`.
- `NativeElementBridge` parses and diffs each valid incoming tree, then assigns the result to `NativeUIBridge.currentTree.value`.
- The differ can legitimately reuse the previous root, while the remaining tree metadata also remains unchanged.
- Compose's default structural equality policy suppresses the assignment when the resulting `NativeUITree` compares equal to the current value.

## Why iOS behaves differently
The iOS bridge exposes its tree as an `@Published` property. Combine publishes on every assignment, so a subscriber can observe an authoritative tree publication even when the assigned value is equal to the preceding value.

Android's `mutableStateOf` models state changes rather than assignment events. Its default structural equality policy therefore cannot, by itself, represent the publication contract.

## Expected behavior
Plugin renderers can observe every successfully mounted tree publication, including a publication whose tree payload is equal to the previous payload.

The existing `currentTree` state should continue to represent the latest tree and retain structural equality behavior for normal Compose rendering performance.

## Actual behavior
Plugin renderers can observe only publications that change `currentTree` according to Compose's equality policy. Equal publications are indistinguishable from no publication.

### How to reproduce the bug

1. PHP publishes tree A.
2. A native plugin renderer temporarily changes local UI state while it sends an event to PHP.
3. PHP rejects, normalizes, or otherwise does not accept that tentative value.
4. PHP authoritatively publishes tree A again.
5. NativePHP successfully parses and mounts the publication, but the differ produces a tree equal to the existing `currentTree` value.
6. Compose suppresses the equal assignment.
7. A plugin observing `currentTree` receives no signal and cannot reconcile its temporary state back to the server-authoritative value.

The same failure can be reduced to publishing the exact same `NativeUITree` instance twice: the core has processed two publications, while an observer sees only the first.

### Debug Output

n/a

macOS

Android (Emulator)

### Notes

_No response_

wojt-janowski:
Resolves #279 

## Summary
- expose a monotonically increasing Android tree-publication revision;
- route every successfully mounted native tree through one publication boundary;
- preserve the existing structural equality behavior of `currentTree`;
- cover repeated publication of the exact same tree with a focused unit test.

## Problem
Android currently exposes the latest native UI tree through a Compose `MutableState<NativeUITree?>`:

```kotlin
val currentTree = mutableStateOf<NativeUITree?>(null)
```

Compose's default state policy suppresses an assignment when the incoming tree compares equal to the current value. That behavior is correct for consumers interested only in changed state, but it means a consumer cannot observe that NativePHP successfully processed a new authoritative publication whose payload is equal to the previous tree.

An equal publication can still be meaningful. For example, a native control may have temporarily updated local state while an event was sent to PHP. If PHP rejects or normalizes that event and republishes the unchanged authoritative tree, the control needs to observe the publication so it can reconcile. Today, the Android bridge makes that publication indistinguishable from no publication.

The equivalent iOS bridge uses `@Published`, whose subscribers can observe each assignment, including an equal value.

## Solution

Add a public, observable `treePublicationId` and an internal `publishTree` boundary:

```kotlin
val currentTree = mutableStateOf<NativeUITree?>(null)
val treePublicationId = mutableLongStateOf(0L)

internal fun publishTree(tree: NativeUITree) {
    currentTree.value = tree
    treePublicationId.longValue++
}
```

`NativeElementBridge` now calls `publishTree(diffedTree)` after a tree has been successfully parsed, diffed, and posted to the main thread.

Consumers that need assignment/publication semantics can observe `treePublicationId` and then read `currentTree`. Existing consumers that only need changed-tree semantics can continue observing `currentTree` unchanged.

## Why a separate revision?

Changing `currentTree` to `neverEqualPolicy()` would force every equal tree assignment to invalidate all Compose consumers of the tree. That would turn a targeted publication requirement into a global recomposition policy and discard the useful structural equality behavior already present.

The explicit revision keeps the two contracts separate:

- `currentTree` is the latest tree state;
- `treePublicationId` records that another authoritative tree was mounted.

The publication method is internal so plugins can observe the revision but cannot manufacture core tree publications.

## Behavior preserved

- `currentTree` retains its existing structural equality policy.
- Existing tree-diff and subtree-reuse behavior is unchanged.
- Existing screen navigation revision behavior is unchanged.
- Failed tree parses do not advance the publication revision.
- Stopping the native UI system and clearing `currentTree` do not masquerade as mounted tree publications.
- No iOS code or public event payload is changed.

## Test plan

- [x] Add a unit test that publishes the exact same `NativeUITree` instance twice.
- [x] Assert that `currentTree` still contains that tree.
- [x] Assert that an observer sees the initial revision and both subsequent increments.
- [x] Run the focused unit test from a generated Android host.
- [x] Run the complete Android JVM unit-test task from a fresh rerun.
- [x] Run `git show --check` on the resulting commit.

Commands:

```bash
./gradlew testDebugUnitTest --tests com.nativephp.mobile.ui.nativerender.NativeUIBridgeTest
./gradlew testDebugUnitTest --rerun-tasks
git show --check HEAD
```

Results:

- Focused regression: `BUILD SUCCESSFUL`.
- Full Android JVM unit suite: `BUILD SUCCESSFUL`; 2 tests, 0 failures.
- Patch whitespace/error check: passed.

## Scope

This change adds the missing Android publication signal only. It does not change tree equality, tree diffing, navigation transitions, event dispatch, or renderer state policy.

### DIFF
--- a/resources/androidstudio/app/src/main/java/com/nativephp/mobile/ui/nativerender/NativeElementBridge.kt
+++ b/resources/androidstudio/app/src/main/java/com/nativephp/mobile/ui/nativerender/NativeElementBridge.kt
@@ -374,7 +374,7 @@ class NativeElementBridge private constructor() {
                             // animation with a slide overlay.
                             if (isFreshStackMount) NavigationCoordinator.reset()
                             if (isNav && !nativeChromeContinuation) NativeUIBridge.screenKey.intValue++
-                            NativeUIBridge.currentTree.value = diffedTree
+                            NativeUIBridge.publishTree(diffedTree)
                             // First publish after a hot-reload dismisses
                             // the "Reloading…" pill and clears the
                             // tree-preservation flag. Both are set by
--- a/resources/androidstudio/app/src/main/java/com/nativephp/mobile/ui/nativerender/NativeUIBridge.kt
+++ b/resources/androidstudio/app/src/main/java/com/nativephp/mobile/ui/nativerender/NativeUIBridge.kt
@@ -2,6 +2,7 @@ package com.nativephp.mobile.ui.nativerender
 
 import android.util.Log
 import androidx.compose.runtime.mutableIntStateOf
+import androidx.compose.runtime.mutableLongStateOf
 import androidx.compose.runtime.mutableStateOf
 
 /**
@@ -18,6 +19,14 @@ class NativeUIBridge private constructor() {
 
         /** Current parsed tree — observed by the Compose renderer */
         val currentTree = mutableStateOf<NativeUITree?>(null)
+
+        /** Public revision for observing every mounted tree publication, including equal trees. */
+        val treePublicationId = mutableLongStateOf(0L)
+
+        internal fun publishTree(tree: NativeUITree) {
+            currentTree.value = tree
+            treePublicationId.longValue++
+        }
 
         /** Whether the native UI system is active */
         val isActive = mutableStateOf(false)
--- a/resources/androidstudio/app/src/test/java/com/nativephp/mobile/ui/nativerender/NativeUIBridgeTest.kt
+++ b/resources/androidstudio/app/src/test/java/com/nativephp/mobile/ui/nativerender/NativeUIBridgeTest.kt
@@ -1,0 +1,53 @@
+package com.nativephp.mobile.ui.nativerender
+
+import androidx.compose.runtime.snapshotFlow
+import androidx.compose.runtime.snapshots.Snapshot
+import kotlinx.coroutines.CoroutineStart
+import kotlinx.coroutines.flow.take
+import kotlinx.coroutines.flow.toList
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.yield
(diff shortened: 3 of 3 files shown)


simonhamp (approved):
LGTM

PR #281, opened 2026-08-04
Fix deep link delivery in native-ui (EDGE) apps on iOS

shanerbaner82:
## The bug

Native-ui apps never call `markWebViewReady()` — the WebView is only the `php://` transport and no page-load callback ever fires — so `DeepLinkRouter`'s `isWebViewReady && isPhpReady` gate parks **every warm deep link as "pending" forever**.

The headline casualty: **all OAuth flows in SuperNative apps are dead on arrival.** `Browser::auth()` completes, the custom-scheme callback reaches `DeepLinkRouter.handle()`, normalization succeeds… and the route is stored and never delivered:

```
🔗 DeepLinkRouter.handle() called with: galactica://auth/callback/12%7CETAb…
🔗 Current state - WebView ready: false, PHP ready: true
🔗 Normalized to: php://127.0.0.1/auth/callback/12|ETAb…
🔗 Not ready, storing as pending URL        ← dropped forever
```

## The fix

- **Warm path**: gate on `isPhpReady && (isWebViewReady || NativeUIBridge.shared.isActive)`. The native-ui branch dispatches through `NativeElementBridge.sendNativeEvent("__deeplink", …)` and never touches the WebView — PHP readiness is all it needs. Dispatch extracted into `dispatchNativeUIDeepLink()`.
- **Pending replay**: same gating; in native-ui mode it replays through the `__deeplink` event instead of `redirectToURL()`, whose WebView load would queue behind the blocked native-ui loop.
- **Cold-start race**: `NativeUIBridge.isActive` gained a `didSet` calling the new `DeepLinkRouter.replayPendingDeepLink()` — a launch-by-deep-link arrives before native-ui takes over, `markPhpReady` fires first, and the replay would otherwise no-op with the link stuck.

Android is unaffected: `MainActivity.navigateWarm()` already gates on `bootReady` alone and dispatches `__deeplink` when native-ui is active.

## Verified

End-to-end GitHub OAuth in the iOS simulator: `Browser::auth()` → ASWebAuthenticationSession → custom-scheme callback → `/auth/callback/{token}` native route mounts and stores the token. Also exercised via `xcrun simctl openurl` warm dispatch.

Two adjacent traps we hit while diagnosing (worth docs mentions, not part of this PR): the ASWebAuthenticationSession fallback callback scheme is the literal string `native`, while docs examples imply `nativephp://`; and OAuth tokens containing `|` (Sanctum's format) must be percent-encoded or iOS's URL parser rejects the callback.


shanerbaner82:
Opened from a local branch that included unpushed WIP — closing; a clean single-commit PR will follow.

PR #289, opened 2026-08-04
Env-gated plugin feature bundles

shanerbaner82:
## What

Plugins can now ship optional pieces of themselves behind an environment variable, so an app compiles in only the SDK surface it actually uses:

```json
"features": {
    "crashlytics": {
        "env": "NATIVEPHP_FIREBASE_CRASHLYTICS",
        "bridge_functions": [ ... ],
        "ios": {"dependencies": {"swift_packages": [{"url": "…firebase-ios-sdk", "products": ["FirebaseCrashlytics"]}]}},
        "android": {"dependencies": {"implementation": [ ... ]}, "gradle_plugins": [ ... ]}
    }
}
```

The motivating case is `nativephp/mobile-firebase`, which is growing Crashlytics, App Check, Performance, and In-App Messaging — nobody should link all of Firebase to send a push.

## How

Resolution happens in `PluginManifest::__construct`, before the readonly sections are assigned: enabled bundles merge into `bridge_functions`, `events`, `ios`, and `android`. **Every downstream consumer — both compilers, `plugin:list`, all `Plugin` accessors — sees one effective manifest and needed no feature awareness.** Lists concatenate, maps (info_plist, entitlements) merge key-wise.

A disabled bundle leaves no trace: no SDK products, no Gradle coordinates, and no bridge registrations pointing at classes that were never compiled.

Two things worth calling out for review:

1. **Feature sources live at `resources/features/<name>/<platform>/`, not under `resources/ios|android`.** Those roots are copied *recursively* (`copySwiftFilesRecursively`, `allFiles`), so a subdirectory would compile in regardless of the gate — and reference products the build never linked. `Plugin::getFeatureSourcePaths()` returns only enabled dirs (overridable per feature with `source_dir`); both compilers copy them alongside the plugin's own sources.
2. **Swift packages need no merging.** `injectSwiftPackage` already resolves a repeated package url to the existing `XCRemoteSwiftPackageReference` and attaches the extra products to the target, so a feature declares the same url carrying only what it needs. (I wrote a URL-union merge first and deleted it — core had this covered.)

Backward compatible: a manifest without `features` is unchanged in every path, and `env` is optional (a bundle without one is always on, useful purely for grouping).

## Tests

New `tests/Unit/Plugins/PluginFeatureGateTest.php` (8 tests): disabled leaves no trace, enabled merges every section, shared-package products, falsy env values (`false`/`0`/``/null), no-env bundles, key-wise map merging, manifests without features, and source paths appearing only when gated on.

Full suite green: 446 unit (2 skipped, 2 risky), 127 plugin-feature. Pint clean.

### DIFF
--- a/src/Plugins/Compilers/AndroidPluginCompiler.php
+++ b/src/Plugins/Compilers/AndroidPluginCompiler.php
@@ -445,14 +445,23 @@ class AndroidPluginCompiler
     {
         $sourcePath = $plugin->getAndroidSourcePath();
 
-        if (! $this->files->isDirectory($sourcePath)) {
+        // Sources from enabled feature bundles only — a disabled feature's
+        // Kotlin never reaches the project, so it can't reference SDK
+        // artifacts the build didn't add.
+        $sourcePaths = $this->files->isDirectory($sourcePath) ? [$sourcePath] : [];
+        $sourcePaths = array_merge($sourcePaths, $plugin->getFeatureSourcePaths('android'));
+
+        if ($sourcePaths === []) {
             return;
         }
 
         $javaBasePath = $this->androidProjectPath.'/app/src/main/java';
 
         // Copy all Kotlin files
-        $files = $this->files->allFiles($sourcePath);
+        $files = array_merge(...array_map(
+            fn (string $path): array => $this->files->allFiles($path),
+            $sourcePaths,
+        ));
 
         foreach ($files as $file) {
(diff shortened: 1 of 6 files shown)


simonhamp:
#262 introduced some changes to the recursive copy. Can we incorporate some flavour of that into this `features` feature so that we don't end up with the same problem?

PR #293, opened 2026-08-05
Fix native:check-build-number for iOS

spaantje:
The iOS path reads --api-key-path, --api-key-id and --api-issuer-id, but CheckBuildNumberCommand only declares --api-key. Symfony therefore throws `The "api-key-path" option does not exist` inside getLatestIosBuildNumber(), where the catch-all turns it into a warning and returns null. The command then prints "No releases found or check not implemented" and "Suggested next: 1" — which is indistinguishable from a brand-new app, so the number cannot be trusted and the command is unusable for the one thing it exists to do.

PackageCommand and BuildIosAppCommand declare all three, which is why the same lookup works there.

Three changes:

- Declare the missing options, matching PackageCommand's wording.
- Resolve credentials from config('nativephp.app_store_connect.*') as a last fallback. getApiKeyPath() only fell back to APP_STORE_API_KEY_PATH, while the published config exposes the key as app_store_connect.api_key (APP_STORE_API_KEY), so a documented setup still found nothing.
- Say "not checked - no API credentials" instead of suggesting a number when the lookup cannot run. Suggesting 1 is worse than declining to answer. Once Apple has actually been asked, "no builds yet for version X" is correct and 1 really is the next build, since build numbers only have to be unique within a version.

### DIFF
--- a/src/Commands/CheckBuildNumberCommand.php
+++ b/src/Commands/CheckBuildNumberCommand.php
@@ -19,6 +19,9 @@ class CheckBuildNumberCommand extends Command
         {platform : The platform to check (android/a, ios/i, or both)}
         {--google-service-key= : Path to Google Service Account JSON key file (Android)}
         {--api-key= : Path to App Store Connect API key file (iOS)}
+        {--api-key-path= : Path to App Store Connect API key file (.p8) - same as --api-key}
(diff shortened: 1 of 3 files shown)


simonhamp (approved):
Looks good. Thanks

PR #295, opened 2026-08-06
Use destructive theme color for native errors

wojt-janowski:
## Summary

- resolve `@nativeError`'s default color from the appearance-aware `destructive` theme token
- preserve explicit color overrides and the existing fallback when the token is unavailable
- add Blade rendering regression coverage

## Why

The directive hardcoded `#FF0000`, so default validation errors ignored each app's light and dark palettes.

## Test plan

- `php -d error_reporting=8191 vendor/bin/pest --compact`
- `vendor/bin/phpstan analyse --no-progress`
- `vendor/bin/pint --test src/NativeServiceProvider.php tests/Feature/Edge/BladeNativeRenderingTest.php`

### DIFF
--- a/src/NativeServiceProvider.php
+++ b/src/NativeServiceProvider.php
@@ -436,11 +436,13 @@ class NativeServiceProvider extends PackageServiceProvider
             return "<?php
                 \$__nativeErrorArgs = [{$expression}];
                 \$__nativeErrorField = \$__nativeErrorArgs[0];
-                \$__nativeErrorColor = \$__nativeErrorArgs[1] ?? theme('destructive', '#FF0000');
+                \$__nativeErrorColor = \$__nativeErrorArgs[1] ?? config('native-ui.theme.light.destructive', '#FF0000');
+                \$__nativeErrorDarkColor = isset(\$__nativeErrorArgs[1]) ? null : config('native-ui.theme.dark.destructive');
                 if (isset(\$errors) && is_array(\$errors) && !empty(\$errors[\$__nativeErrorField])) {
                     \\Native\\Mobile\\Edge\\NativeElementCollector::leaf('text', [
                         'text' => \$errors[\$__nativeErrorField],
                         'color' => \$__nativeErrorColor,
+                        'dark' => ['color' => \$__nativeErrorDarkColor],
                         'fontSize' => 12,
                     ]);
                 }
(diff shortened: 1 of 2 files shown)


shanerbaner82 (changes requested):
Right idea — the hardcoded `#FF0000` should absolutely come from the `destructive` token, and the override/fallback handling here is correct. But the implementation resolves the mode at render time, which breaks live appearance switching, so I'd like one more pass before this lands.

**The problem:** `theme('destructive', …)` reads `System::isDarkMode()` when the view renders and bakes a single hex into the tree. EDGE deliberately does *not* re-render on appearance changes — the only `AppearanceChanged` listener in core refreshes the appearance cache, and the rendered tree stays alive while the native side swaps `dark_*` companion props. That's how every other theme consumer works: `text-theme-*` / `bg-theme-*` resolve both palettes and emit a dark companion.

Concretely: render an error in light mode → text is `#B91C1C`. User flips to dark → everything else swaps via its `dark_*` props, but the error stays `#B91C1C` — the exact color the dark palette overrides to `#F87171` because it's illegible on dark backgrounds — until something forces a re-render. So the fix is appearance-aware at initial render only.

**Suggested change:** the directive's `leaf()` path already flows through `buildDarkProps()`, so a `'dark' => ['color' => …]` attr becomes a live-swapping `dark_color` prop. Read both palettes from config directly:

```php
$__nativeErrorColor = $__nativeErrorArgs[1] ?? config('native-ui.theme.light.destructive', '#FF0000');
$__nativeErrorDarkColor = isset($__nativeErrorArgs[1]) ? null : config('native-ui.theme.dark.destructive');
```

…and append `'dark' => ['color' => $__nativeErrorDarkColor]` to the leaf attrs when it's non-null. This also removes the `System::isDarkMode()` bridge dependency from a compiled view entirely. Keeping the companion out of the explicit-override path preserves today's behavior for `@nativeError('email', '#123456')`.

**Test:** setting both `light.destructive` and `dark.destructive` to the same `#B42318` means the test passes no matter which palette the directive reads — it can't tell render-time-light from render-time-dark from companion-based. Please use two different hexes and assert both `props.color` (light value) and `props.dark_color` (dark value).

Everything else checks out locally: full suite green, PHPStan matches main's baseline, Pint clean, and web rendering is unaffected (`$errors` is a MessageBag there, so the directive still no-ops).

wojt-janowski:
Amended as requested:

- Native errors now emit light color and dark dark_color, enabling live device appearance switching without rerendering.
- Explicit custom colours remain fixed across both modes.
- Tests assert distinct light/dark values and custom-colour behavior.

shanerbaner82 (approved):
Follow-up addresses the live appearance-switching concern. Light and dark destructive tokens are emitted separately, explicit overrides remain unchanged, and regression coverage now verifies both companion colors. Looks good.

PR #296, opened 2026-08-06
Add Tailwind aliases and unsupported class warnings

wojt-janowski:
## Summary

- support Tailwind's current `grow`, `grow-0`, `shrink`, and `shrink-0` utilities
- warn once per Blade view in debug mode when unsupported classes are dropped
- preserve diagnostics across parser cache hits and ignore inactive platform variants

Unsupported classes currently return no parsed attributes and are skipped silently, making migration layout regressions difficult to trace.

## Test plan

- `php -d error_reporting=8191 vendor/bin/pest --compact`
- `composer analyse -- --no-progress`
- `vendor/bin/pint --test`
- `git diff --check`

### DIFF
--- a/src/Edge/NativeComponent.php
+++ b/src/Edge/NativeComponent.php
@@ -1296,11 +1296,16 @@ abstract class NativeComponent
         // returns a View. We need the View to access its engine + path.
         $view = view($name, $data);
         $engine = $view->getEngine();
+        TailwindParser::beginViewDiagnostics($view->getName());
 
         if (! $engine instanceof CompilerEngine) {
             // Non-blade engine — fall back to the standard render path.
             // $this won't be bound, but at least the view still runs.
-            $view->render();
+            try {
+                $view->render();
+            } finally {
+                TailwindParser::endViewDiagnostics();
+            }
 
             return;
         }
@@ -1373,6 +1378,7 @@ abstract class NativeComponent
             $factory->flushStateIfDoneRendering();
         } finally {
             NativeTagPrecompiler::setActive($wasActive);
+            TailwindParser::endViewDiagnostics();
         }
     }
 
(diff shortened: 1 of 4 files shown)


shanerbaner82 (approved):
Nice work — the aliases are exactly right, and the diagnostics plumbing is careful where it needs to be. Verified locally: full suite green (risky/skipped counts match main), PHPStan errors identical to main's baseline, Pint clean. Things I checked specifically:

- Variant recursion goes through the private `parseClass`, so `dark:truncate` / `ios:truncate` report once as themselves — no double-reporting.
- `$unsupportedCache` stays in lockstep with `$cache` (all three setters now route through `clearCache()`), and cache hits replay diagnostics into the active scope.
- Both render paths in `renderBladeBoundToSelf` end the diagnostic scope in a `finally`.
- Warn-once genuinely survives `Native::test`'s per-render `clearCache()` — the feature test covers what it claims.

I pushed one small fix to the branch (9af9d58): the feature test wrote fixtures into `tests/Feature/Edge/views/` without creating the directory, so it failed when run standalone (`vendor/bin/pest tests/Feature/Edge/TailwindParserWarningsTest.php` on a fresh clone) and only passed in full runs because an alphabetically earlier test had already made the dir. Now it mkdirs like its siblings.

Non-blocking notes, fine as follow-ups:

1. `$reportedUnsupportedByView` has no reset path — `clearCache()` deliberately doesn't touch it (clearing it there would break the warn-once test), but that means a view name that has warned once in a test process can never warn again. A dedicated `resetDiagnostics()` wired into the test base setup would remove the order-dependence for future tests.
2. Programmatic classes (`Element::class('...')`, e.g. `NativeLayout::bottomBar()`) parse outside any view scope, so unsupported classes there still drop silently. Fine to scope this to Blade — worth a docblock line or follow-up issue.
3. Views shared between web and native are full of `hover:`/`md:`/`transition` etc., which will all be flagged. Debug-only and one log line per view, so acceptable — but a docs note that web-only utilities are reported too would preempt questions.

PR #301, opened 2026-08-08
EDGE: Laravel-style validation for components

shanerbaner82:
## Summary

Adds first-class validation to `NativeComponent`, identical on device and web:

- `#[Validate('required|email')]` on public props + `$this->validate()` / `validateOnly()` in handlers. Rule sources merge: attributes → `rules()` method → inline argument. `validate(SomeFormRequest::class)` harvests a FormRequest's rules/messages/attributes (harvesting only — `authorize()` is never called).
- Failure throws Laravel's own `ValidationException`; `dispatch()` and `dispatchNativeEvent()` route through a `runGuarded()` wrapper that records errors on the component's bag and aborts the handler. Both device run loops and the web target enter through these two methods, so semantics are identical by construction.
- `$errors` (a `ViewErrorBag`) is injected into every view-data merge — `@error('field')` Blade muscle memory just works. The pre-existing `@nativeError` directive accepts the bag too (it previously required a hand-rolled array and silently rendered nothing otherwise).
- Attribute rules are **eager**: `__syncProperty()` runs `validateOnly()` after the `updated` hook, so `native:model`'s live/blur/debounce modifiers become the real-time validation cadence for free. `rules()`-method rules stay on-demand.
- **Error display is explicit-only (Livewire semantics)**: a failed validation records errors and exposes `$errors`, but nothing renders until the author writes `@error`/`@nativeError` or sets `error`/`supporting` attributes on an element. No automatic injection — devs own error presentation entirely. (`compileNativeModel()` emits a `model-prop` metadata attribute, stripped in the collector; it drives nothing.)

## Verification

15-case Pest suite driving the real dispatch cycle via `Native::test()`: handler abort with state kept, success path, inline-rule scoping, FormRequest harvesting with custom messages, author-thrown `withMessages`, eager sync validation (fires/clears/on-demand tier stays quiet), wildcard `validateOnly`, `@error` + `@nativeError` display, child-component bag scoping, and explicit-only display (failed validation leaves bound elements' props untouched). Also exercised end-to-end through the mobile-web target against a demo app.

`TestableComponent::guard()` routes `set()`/`call()` through `runGuarded` so harness interactions get the same validation-abort semantics as real dispatch.

### DIFF
--- a/src/Attributes/Validate.php
+++ b/src/Attributes/Validate.php
@@ -1,0 +1,29 @@
+<?php
+
+namespace Native\Mobile\Attributes;
+
+use Attribute;
+
+/**
+ * Declares validation rules for a public component property.
+ *
+ *     #[Validate('required|email')]
+ *     public string $email = '';
+ *
+ *     #[Validate(['required', 'min:8'])]
+ *     public string $password = '';
+ *
+ * Attribute rules are EAGER: they run automatically whenever the
+ * property syncs through `native:model` (validateOnly inside
+ * __syncProperty), and they're included in every $this->validate()
+ * call. Rules that should only run on demand belong in the component's
+ * rules() method instead — that's the opt-out lever, mirroring the
+ * attribute-vs-method split Livewire settled on.
+ */
+#[Attribute(Attribute::TARGET_PROPERTY)]
+class Validate
+{
+    public function __construct(
+        public string|array $rule,
+    ) {}
+}
--- a/src/Attributes/Validate.php
+++ b/src/Attributes/Validate.php
@@ -19,8 +19,18 @@ use Attribute;
  * call. Rules that should only run on demand belong in the component's
  * rules() method instead — that's the opt-out lever, mirroring the
  * attribute-vs-method split Livewire settled on.
+ *
+ * Repeatable — stacked attributes merge their rules:
+ *
+ *     #[Validate('required')]
+ *     #[Validate('min:3')]
+ *     public string $title = '';
+ *
+ * String rules are pipe-exploded when merging, so use the array form
+ * for rules that legitimately contain a pipe (regex:) — same caveat as
+ * Laravel's validator itself.
  */
-#[Attribute(Attribute::TARGET_PROPERTY)]
+#[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
 class Validate
 {
     public function __construct(
--- a/src/Edge/Concerns/ValidatesProps.php
+++ b/src/Edge/Concerns/ValidatesProps.php
@@ -109,12 +109,26 @@ trait ValidatesProps
         $validator = $this->makePropValidator($rules, $messages, $attributes);
 
         if ($validator->fails()) {
-            $this->nativeErrorBag = new MessageBag($validator->errors()->messages());
+            // Failure refreshes exactly the keys THIS validation ran
+            // (wildcard-aware) and leaves the rest of the bag alone —
+            // never a whole-bag replace. With per-tier event guards,
+            // multiple listeners in one delivery can each fail; a
+            // replace here would erase an earlier tier's recorded
+            // errors (the second failing ->on() sibling wiping the
+            // first). Scoped-refresh also means validate(['bio' => …])
+            // failing doesn't clear an unrelated on-screen title error
+            // it never re-checked.
+            $this->forgetErrors(array_keys($rules));
+            foreach ($validator->errors()->messages() as $key => $keyMessages) {
+                foreach ($keyMessages as $message) {
+                    $this->getErrorBag()->add($key, $message);
+                }
+            }
 
             throw new RecordedValidationException($validator);
         }
 
-        // A full pass supersedes everything previously recorded,
+        // A full PASS still supersedes everything previously recorded,
         // including addError() entries — same as Livewire.
         $this->nativeErrorBag = new MessageBag;
 
--- a/src/Edge/Concerns/ValidatesProps.php
+++ b/src/Edge/Concerns/ValidatesProps.php
@@ -3,6 +3,7 @@
 namespace Native\Mobile\Edge\Concerns;
 
 use Illuminate\Contracts\Validation\Validator as ValidatorContract;
+use Illuminate\Foundation\Http\FormRequest;
 use Illuminate\Support\Facades\Validator;
 use Illuminate\Support\MessageBag;
 use Illuminate\Support\ViewErrorBag;
@@ -36,12 +37,22 @@ trait ValidatesProps
     protected ?MessageBag $nativeErrorBag = null;
 
     /**
-     * True while a ValidationException thrown by validate()/validateOnly()
-     * has already been recorded in the bag — tells runGuarded() not to
-     * fold the exception's messages in a second time (its per-key merge
-     * would otherwise resurrect keys a later validate() had cleared).
+     * Exceptions whose messages have already been recorded in SOME
+     * component's bag. Keyed on the exception INSTANCE (not an instance
+     * flag) because guards nest across components: a child's guard can
+     * catch an exception a parent's validate() already recorded (emit()
+     * listeners, nested dispatch), and an instance flag on the catcher
+     * would wrongly fold the parent's errors into the child's bag — or,
+     * left stale, wrongly swallow a fresh author-thrown exception.
+     *
+     * @var \WeakMap<ValidationException, true>
      */
-    private bool $validationRecorded = false;
+    private static ?\WeakMap $recordedValidationExceptions = null;
+
+    private static function recordedExceptions(): \WeakMap
+    {
+        return self::$recordedValidationExceptions ??= new \WeakMap;
+    }
 
     // ── Bag access ──────────────────────────────────
 
@@ -115,9 +126,11 @@ trait ValidatesProps
 
         if ($validator->fails()) {
             $this->nativeErrorBag = new MessageBag($validator->errors()->messages());
-            $this->validationRecorded = true;
 
-            throw new ValidationException($validator);
+            $exception = new ValidationException($validator);
+            self::recordedExceptions()[$exception] = true;
+
+            throw $exception;
         }
 
         // A full pass supersedes everything previously recorded,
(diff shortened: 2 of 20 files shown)


src/Attributes/Validate.php
```diff
@@ -0,0 +1,29 @@
...
+ * rules() method instead — that's the opt-out lever, mirroring the
+ * attribute-vs-method split Livewire settled on.
+ */
+#[Attribute(Attribute::TARGET_PROPERTY)]
+class Validate
+{
+    public function __construct(
+        public string|array $rule,
+    ) {}
+}
```
shanerbaner82, line 39: **Must-fix: `#[Validate]` isn't repeatable and fatals if stacked.**

`#[Attribute(Attribute::TARGET_PROPERTY)]` omits `IS_REPEATABLE`. Livewire's `Validate` attribute *is* repeatable, so the natural

```php
#[Validate('required')]
#[Validate('min:3')]
public string $title;
```

is exactly what authors will write — and `attributeValidationRules()` iterates `$prop->getAttributes(Validate::class)` and calls `newInstance()`, which throws `Error: Attribute ... must not be repeated` on the first render or keystroke, crashing the whole screen. The harvesting `foreach` is even written to handle multiple attributes (last wins), which implies repetition was expected to work.

Fix: add `Attribute::IS_REPEATABLE` and *merge* the rules (don't let last-wins drop the earlier ones) — or reject at boot with a clear message.

src/Edge/NativeComponent.php
```diff
@@ -292,7 +294,7 @@ protected function renderStreaming(): bool
 
     protected function view(string $name, array $data = []): Element
     {
-        $viewData = array_merge($this->getPublicProperties(), $data);
+        $viewData = array_merge(['errors' => $this->errorBagForViews()], $this->getPublicProperties(), $data);
```
shanerbaner82: **Legacy `public $errors` support is an accident of `array_merge` ordering.**

Any component declaring an `$errors` prop — including the documented v3 default `public array $errors = []` — permanently shadows the injected `ViewErrorBag` here. If that component adopts `$this->validate()` or `#[Validate]`, validation records to the trait's bag but Blade sees the stale empty array: `@nativeError` renders nothing and Laravel's `@error` fatals with `Call to a member function getBag() on array`. Only the exact tested shape (non-empty array + `@nativeError` only, per `LegacyErrorsScreen`) works.

Suggest making the policy explicit instead of encoded in merge precedence: reserve the property name at boot (Livewire-style reserved-props check) or an explicit `property_exists` fallback with a deprecation notice.

src/Edge/NativeComponent.php
```diff
@@ -334,7 +336,7 @@ protected function view(string $name, array $data = []): Element
      */
     protected function partial(string $name, array $data = []): Element
     {
-        $viewData = array_merge($this->getPublicProperties(), $data);
+        $viewData = array_merge(['errors' => $this->errorBagForViews()], $this->getPublicProperties(), $data);
```
shanerbaner82: **The `array_merge(['errors' => ...], $this->getPublicProperties(), $data)` envelope is copy-pasted into five render methods** (`view`, `partial`, `fromView`, `fromViewPartial`, `streamView`). Three independent review angles converged on this.

Nothing enforces "every render path injects `$errors`": the next render method needs five identical edits, and a missed one is a path where `@error` fatals with undefined `$errors` only when a template happens to render through it. It's also the single right home for the legacy-`$errors` fallback policy (see comment at line 297) instead of encoding it five times in merge order.

Extract one `protected function viewData(array $data = []): array` seam.

shanerbaner82 (commented):
Second-round review of the revised head (`730d611`).

**All 10 first-round findings verified genuinely fixed** — eager sync now passes attribute rules explicitly, child-emit guards run on the owner, the WeakMap replaces the stale instance flag, wildcard sibling scoping is fixed, FormRequest `rules()` gets container method-injection, per-keystroke reflection is cached, and both test-harness divergences are resolved. Nice work.

This round surfaced **10 new findings**, mostly in code the fixes touched or entry points the first round didn't cover. Eight are inline comments below; two sit outside the diff hunks so they're here:

---

**1. (Must-fix) `#[Poll]` methods and `onBackPressed()` can kill the entire runloop** — `src/Edge/NativeComponent.php:1621` and `:2389`.

These author-code entry points run outside `runGuarded` **and** outside any try/catch. Now that this PR makes `$this->validate()` routine control flow ("abort handler, keep state, render with errors"), an author calling it in a `#[Poll]`-ed method (periodic revalidate/auto-save) or in `onBackPressed()` ("validate before leaving") throws a `ValidationException` that escapes `runLoop()` entirely — `NativeRouter::loop()` only wraps `mount()`/`onResume()`, so the PHP runloop dies: frozen app, not even the error screen. Unlike `dispatch()`/render, there's no `catch (\Throwable)` → `renderErrorScreen` at either call site.

Deeper fix worth considering: one handler-invocation seam that every entry point (dispatch, native events, child events, polls, back, test `set()`/`call()`) routes through — several findings below are symptoms of not having that seam.

---

**2. Harness/device guard mismatch in `search()` and `call()`, in both directions** — `src/Testing/TestableComponent.php:567`.

`search()` wraps `onSearchQuery` in plain `guard()`, so a validating `onSearchQuery()` throws raw `ValidationException` in tests — but on device, `search_query` routes through `dispatch()`'s `runGuarded` and is recorded-and-swallowed. The harness certifies a crash the device doesn't have. Mirror image: `->call('mount')` (or any non-handler method) routes through `guardDispatch` and quietly records errors, while on device a `mount()`-time failure bubbles to the error screen — this PR's own `MountValidationScreen` docblock says mount failures "must BUBBLE, never be silently recorded". `search()` should use `guardDispatch()`; `call()` shouldn't guard methods that aren't dispatch handlers (or the seam belongs on `NativeComponent` itself, per the point above).

---

Severity read: this finding #1 and the `IS_REPEATABLE` fatal are must-fix crashes; the merge-ordering legacy `$errors` shadow, the four-tier event guard, the harness mismatch above, and the `validateOnly()` double-run are worth fixing before merge since they're all in seams this PR creates. The remaining four (model-prop leak, `@nativeError` falsy collapse, the five-site view-data envelope, the WeakMap) could be follow-ups.

Also verified and deliberately **not** raised (refuted with evidence): web error-bag persistence (the web plugin seals/restores the bag at `WebScreenRunner.php:632/198`) and the bag-wipe semantics (exact Livewire parity, verified against Livewire source).

src/Edge/NativeComponent.php
```diff
@@ -1734,8 +1748,15 @@ public function registerCleanup(\Closure $callback): void
 
     /**
      * Handle a native event (type 20) by looking up #[OnNative] listeners.
+     * Guarded like dispatch(): a ValidationException in a listener
+     * records errors and aborts that listener instead of crashing.
      */
     protected function dispatchNativeEvent(array $event): void
```
shanerbaner82, line 1790: **One `runGuarded` wraps ALL four native-event delivery tiers, so a `ValidationException` in tier 1 silently skips tiers 2–4.**

A native event delivers through four sequential tiers inside `dispatchNativeEventUnguarded`: `fireNativeCallback` (fluent `then()`/`catch()`), `->on()` closure listeners, `dispatchGloballyIfMarked` (Laravel `event()`), then the `#[On]` method. With the guard out here, `Camera::getPhoto()->then(fn () => $this->validate([...]))` failing validation silently skips the component's `->on()` closures, the global broadcast, AND its `#[On(PhotoTaken::class)]` listener for that event — partial delivery with only a bag entry as evidence. The docblock's claim that it "aborts that listener" is wrong by three tiers.

Same guard placement also means a validating `onSearchQuery()` handler invisibly keeps stale results: native search chrome renders no Blade, so the recorded errors have no `@error` outlet at all.

Guard per listener/tier, not per event.

src/Edge/Concerns/ValidatesProps.php
```diff
@@ -0,0 +1,375 @@
...
+                $exception = new ValidationException($validator);
+                self::recordedExceptions()[$exception] = true;
+
+                throw $exception;
+            }
+        }
+
+        $this->forgetErrors([$prop]);
+
+        return $validator->fails() ? [] : $validator->validated();
```
shanerbaner82: **`validateOnly()` runs the entire matched rule set twice, and returns `[]` when only wildcard siblings failed.**

`return $validator->fails() ? [] : $validator->validated();` — Illuminate's `passes()` is not memoized, so this re-executes every rule a second time. This is the eager per-keystroke path: every *passing* keystroke on a `#[Validate('required|email|unique:users')]` prop runs two `unique:` DB queries, and time-dependent rules can even disagree between the two runs.

Separately: with `tags = ['', 'ok']` under `'tags.*' => 'required'`, `validateOnly('tags.1')` correctly declines to throw (only sibling `tags.0` failed) but then returns `[]` — author code doing `$data = $this->validateOnly('tags.1')` gets nothing for a field that passed.

Capture `$failed = $validator->fails()` once, early-return the pass case, and return the target key's validated data in the sibling-only-failure branch.

src/Edge/Concerns/ValidatesProps.php
```diff
@@ -0,0 +1,375 @@
...
+     * component's bag. Keyed on the exception INSTANCE (not an instance
+     * flag) because guards nest across components: a child's guard can
+     * catch an exception a parent's validate() already recorded (emit()
+     * listeners, nested dispatch), and an instance flag on the catcher
+     * would wrongly fold the parent's errors into the child's bag — or,
+     * left stale, wrongly swallow a fresh author-thrown exception.
+     *
+     * @var \WeakMap<ValidationException, true>
+     */
+    private static ?\WeakMap $recordedValidationExceptions = null;
```
shanerbaner82: **The static `WeakMap` is more machinery than the job needs.**

It exists only to distinguish trait-thrown from author-thrown exceptions — but since `runGuarded` never rethrows, each exception instance is seen by at most one guard, so no shared mutable state is required for the cross-component scenarios the docblock describes. Cost: a process-global static in a trait, a lazy-init accessor required only because PHP can't default-init a `WeakMap` property, and a paragraph of instance-keying rationale.

Simpler: have `validate()`/`validateOnly()` throw `RecordedValidationException extends ValidationException`; `runGuarded` folds only when `! $e instanceof RecordedValidationException`. Author catch-blocks (`catch (ValidationException)`) and the catch-then-rethrow regression test behave identically. Fine as a follow-up.

src/Edge/NativeTagPrecompiler.php
```diff
@@ -386,7 +386,13 @@ private function compileNativeModel(string $prop, string $modifiers): string
             // `.live` or anything unknown falls through to syncMode=live.
         }
 
-        $out = ':value="$'.$prop.'" _change="__syncProperty(\''.$prop.'\')" sync-mode="'.$syncMode.'"';
+        // `model-prop` carries the bound property name into the collector
+        // as metadata (stripped there before the element sees attrs) —
+        // useful for devtools/future targets. It deliberately does NOT
+        // drive any automatic error display; errors render only where
```
shanerbaner82: **The new `model-prop` attribute has zero consumers yet leaks past its single strip point.**

`createElement()`'s `unset` (NativeElementCollector:1245) doesn't cover the streaming paths (`applyAttributes` receives it there — core elements allowlist it away, but the collector comment's promise to "strip it before the element sees the attrs" is false for plugins) or the child-component funnel: `applyChildProps`' skip filter only excludes `native-`/`_` prefixes, so `<native:my-child native:model="choice">` forwards `model-prop` → `Str::camel` → clobbers any `public $modelProp` on the child with the string `'choice'` every render.

The plumbing is justified only as "useful for devtools/future targets", and the same information is already derivable from the `_change="__syncProperty('prop')"` expression that `TestableComponent` parses today. Suggest deleting it until a consumer exists — or at minimum stripping at the shared attr funnel.

src/NativeServiceProvider.php
```diff
@@ -433,13 +433,24 @@ protected function registerBladeDirectives(): void
...
+            // hand-rolled ['field' => 'message'] arrays working.
             return "<?php
                 \$__nativeErrorArgs = [{$expression}];
                 \$__nativeErrorField = \$__nativeErrorArgs[0];
                 \$__nativeErrorColor = \$__nativeErrorArgs[1] ?? '#FF0000';
-                if (isset(\$errors) && is_array(\$errors) && !empty(\$errors[\$__nativeErrorField])) {
+                \$__nativeErrorMessage = null;
+                if (isset(\$errors)) {
+                    if (\$errors instanceof \\Illuminate\\Support\\ViewErrorBag) {
+                        \$__nativeErrorMessage = \$errors->first(\$__nativeErrorField) ?: null;
```
shanerbaner82: **`$errors->first($field) ?: null` treats a falsy message string (`'0'`) as no-error.**

`addError('qty', '0')` (or a translation/custom message resolving to `'0'`): `$errors->has('qty')` is true and `@error('qty')` renders it, but `@nativeError('qty')` collapses `'0' ?: null` to `null` and emits nothing — disagreeing with `@error` in the same view. The old array branch's `!empty()` quirk got extended onto the new bag path instead of fixed. Use `has()` / a `!== ''` check.

simonhamp (approved):

PR #302, opened 2026-08-08
Durable fluent native callbacks: $this closures cross the process boundary + onSuccess()

shanerbaner82:
## Summary

`Camera::pickImages('image')->onSuccess(function ($media) { $this->attachedImage = …; })` now fires even when the registering process is gone — an OS kill on device, or the next request on the stateless web target.

- **The insight:** `fireNativeCallback` already rebinds every non-static closure to the *live* component (full class scope) at fire time, so the closure's `$this` binding never needs to survive serialization — only its code and captured `use` vars do. `NativeCallbacks::register()` now rebinds `$this`-bound closures to a throwaway `stdClass` carrier and stores them in the existing durable cache tier via `SerializableClosure` (previously such closures skipped the durable tier entirely).
- `onSuccess()` sugar on `HandlesNativeCallbacks`: names the builder's success event generically (photoTaken / mediaSelected / videoRecorded all read the same); honors `->event()` overrides.
- `fireNativeCallback` also resolves **method-name strings** (`->mediaSelected('onMediaSelected')`) against the live component — a fluent form with zero serialization constraints. Unknown strings are logged and dropped instead of fataling.

Constraints (documented in the code): captured `use` vars must serialize; closures in eval'd code can't round-trip (SerializableClosure extracts source text); the durable tier needs a persistent cache store.

Independent of #301 — applies cleanly to main; the two compose without conflict.

## Verification

Carrier round-trip proven in isolation (serialize → fresh process → rebind: private-property access restored, captures intact), then end-to-end via the mobile-web target against a demo app: closure registered in one HTTP request, fired in a later request on a rebuilt component instance. Package-level tests to follow.

### DIFF
--- a/src/Concerns/HandlesNativeCallbacks.php
+++ b/src/Concerns/HandlesNativeCallbacks.php
@@ -46,12 +46,28 @@ trait HandlesNativeCallbacks
     /**
      * Register a callback for the operation's SUCCESS event without naming
      * it — `->onSuccess(fn ($event) => …)` reads the same on every builder
-     * (photoTaken, mediaSelected, videoRecorded, …). Honors a custom
-     * ->event(Custom::class) override, since that replaces $eventClass.
+     * (photoTaken, mediaSelected, videoRecorded, …). A ->event(Custom::class)
+     * override is honored in EITHER order: event() re-keys registrations
+     * made under the previous success event (see retargetPendingCallbacks).
      */
     public function onSuccess(Closure|array|string $callback): static
     {
         return $this->on($this->eventClass, $callback);
+    }
+
+    /**
+     * Re-key callbacks registered under a previous success event class.
+     * Every builder's event() calls this before assigning the override,
+     * so ->onSuccess(...)->event(Custom::class) doesn't leave the
+     * callback stranded under the default event, silently never firing.
+     */
+    protected function retargetPendingCallbacks(?string $from, string $to): void
+    {
+        if ($from === null || $from === $to) {
+            return;
+        }
+
+        NativeCallbacks::retarget($this->getId(), $from, $to);
     }
 
     /**
--- a/src/Concerns/HandlesNativeCallbacks.php
+++ b/src/Concerns/HandlesNativeCallbacks.php
@@ -35,6 +35,10 @@ trait HandlesNativeCallbacks
     /**
      * Register a callback for a specific event class. Public escape hatch for
      * custom events; also the primitive every named method funnels through.
+     *
+     * Ordering: call ->id() BEFORE registering callbacks — registration
+     * keys on the id at call time, and only ->event() re-keys existing
+     * registrations (a later ->id() does not).
      */
     public function on(string $eventClass, Closure|array|string $callback): static
     {
--- a/src/Concerns/HandlesNativeCallbacks.php
+++ b/src/Concerns/HandlesNativeCallbacks.php
@@ -44,6 +44,17 @@ trait HandlesNativeCallbacks
     }
 
     /**
+     * Register a callback for the operation's SUCCESS event without naming
+     * it — `->onSuccess(fn ($event) => …)` reads the same on every builder
+     * (photoTaken, mediaSelected, videoRecorded, …). Honors a custom
+     * ->event(Custom::class) override, since that replaces $eventClass.
+     */
+    public function onSuccess(Closure|array|string $callback): static
+    {
+        return $this->on($this->eventClass, $callback);
+    }
+
+    /**
      * Event classes that get a named handler method. Defaults to the success
      * event plus any failure events; override to expose more.
      *
--- a/src/Edge/NativeComponent.php
+++ b/src/Edge/NativeComponent.php
@@ -1922,8 +1922,24 @@ abstract class NativeComponent
             return;
         }
 
-        if (is_string($callback) && class_exists($callback)) {
-            $callback = app($callback);
+        if (is_string($callback)) {
+            if (class_exists($callback)) {
+                $callback = app($callback);
+            } elseif (method_exists($this, $callback)) {
+                // Method-name form — `->mediaSelected('onMediaSelected')`.
+                // Serializes into the durable tier trivially, so it's the
+                // fluent shape that survives both a process kill on device
+                // AND the stateless web target, where every request is a
+                // fresh process and closures registered last request are
+                // gone unless the durable tier carried them.
+                $method = $callback;
+                $callback = fn ($event) => $this->{$method}($event);
+            } else {
+                NativeRouter::debugLog("Native callback '{$callback}' is neither a class nor a method on ".static::class);
+                NativeCallbacks::forget($id, $eventClass);
+
+                return;
+            }
         }
 
         // Bind the closure to this live component so then()/catch() can use
--- a/src/Edge/NativeComponent.php
+++ b/src/Edge/NativeComponent.php
@@ -1911,11 +1911,13 @@ abstract class NativeComponent
         // Exact correlation by id (camera). If that misses — either no id came
         // back (some native paths drop it across a lifecycle bounce, e.g. the
         // gallery picker) or it didn't match — fall back to the single in-flight
-        // callback for this event class.
-        $callback = ($id !== null) ? NativeCallbacks::resolve($id, $eventClass) : null;
+        // callback for this event class. `for: $this` lets the registry skip a
+        // durable entry that belongs to a DIFFERENT screen class (post-kill
+        // cold start lands on '/', not the screen that registered).
+        $callback = ($id !== null) ? NativeCallbacks::resolve($id, $eventClass, for: $this) : null;
 
         if ($callback === null) {
-            [$id, $callback] = NativeCallbacks::resolveByEvent($eventClass) ?? [null, null];
+            [$id, $callback] = NativeCallbacks::resolveByEvent($eventClass, for: $this) ?? [null, null];
         }
 
         if ($callback === null) {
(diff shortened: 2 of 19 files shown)


src/Support/NativeCallbacks.php
```diff
@@ -49,13 +49,19 @@ public static function register(string $id, string $eventClass, Closure|array|st
...
+                // the LIVE component (full class scope, so private access
+                // is restored) at fire time. Only the code + captured
+                // `use` vars must cross the boundary, so rebind to a
+                // throwaway serializable carrier first. This is what lets
+                // `->onSuccess(function ($e) { $this->… })` fire on the
+                // stateless web target (fresh process per request) and
+                // survive an OS kill on device.
                 if ((new \ReflectionFunction($serializable))->getClosureThis() !== null) {
-                    return;
+                    $serializable = Closure::bind($serializable, new \stdClass);
```
shanerbaner82: **Must-fix (reproduced): two closures on one source line serialize with the SAME body — after a process death the cancel handler runs the success body.**

serializable-closure v1.3.7 extracts code by start line only (`ReflectionClosure.php:88` uses `getStartLine()`), so

```php
->photoTaken(fn ($e) => $this->save($e))->photoCancelled(fn () => $this->reset())
```

on one line gives **both** durable copies the first closure's code. Warm path is correct (real closures in memory), but after an OS kill the durable copy of the cancel callback executes `save()` — e.g. saves the photo on cancel. No exception anywhere, only after a kill, only for one-line chains: near-undiagnosable in the field. Pre-PR only rare static closures hit the serializer; this carrier routes the dominant `$this`-bound fluent shape through it, so the exposure is new.

Worth at least detecting (two registrations from the same file+line → drop durability for both with a debugLog) if not fixing upstream.

src/Support/NativeCallbacks.php
```diff
@@ -49,13 +49,19 @@ public static function register(string $id, string $eventClass, Closure|array|st
...
+                // serializable. But the binding doesn't need to survive —
+                // fireNativeCallback rebinds every non-static closure to
+                // the LIVE component (full class scope, so private access
+                // is restored) at fire time. Only the code + captured
+                // `use` vars must cross the boundary, so rebind to a
+                // throwaway serializable carrier first. This is what lets
+                // `->onSuccess(function ($e) { $this->… })` fire on the
+                // stateless web target (fresh process per request) and
+                // survive an OS kill on device.
                 if ((new \ReflectionFunction($serializable))->getClosureThis() !== null) {
```
shanerbaner82: **First-class callables lose durability noisily** (verified on PHP 8.4): `->onSuccess($this->onPicked(...))` produces a *fake closure* — `getClosureThis()` is non-null, so it enters this branch, but `Closure::bind` on a method-derived closure emits `Cannot bind method ... to object of class stdClass` and returns **null**; `new SerializableClosure(null)` TypeErrors (or the warning becomes an ErrorException under Laravel), lands in the catch, and `report()`s on every registration — log/Sentry noise per capture, and on the web target the callback never fires.

Ironic gap: SerializableClosure handles fake closures fine *without* the rebind (probe-verified). Detect them — `(new \ReflectionFunction($c))->isAnonymous() === false` — and skip the carrier to recover durability for this natural spelling.

src/Support/NativeCallbacks.php
```diff
@@ -49,13 +49,19 @@ public static function register(string $id, string $eventClass, Closure|array|st
...
-                    return;
+                    $serializable = Closure::bind($serializable, new \stdClass);
                 }
 
                 $serializable = new SerializableClosure($serializable);
```
shanerbaner82, line 175: **Closures capturing raw resources write a POISONED durable copy** (reproduced): `serialize()` doesn't throw on resources — it silently turns them into ints — so `register()`'s try/catch never fires and the durable copy is written corrupt. After a process kill the restored closure throws `TypeError: must be of type resource, int given` mid event-delivery; `fireNativeCallback`'s try/**finally** has no catch, so the runloop paints the component error screen and the `finally` consumes the callback — result permanently lost. The catch's own comment ("Closure captured ... a resource") is factually wrong for resources: they never reach it.

Needs a post-serialize sanity check (round-trip and compare, or scan the use-vars for resources and skip durability with a debugLog).

src/Support/NativeCallbacks.php
```diff
@@ -49,13 +49,19 @@ public static function register(string $id, string $eventClass, Closure|array|st
             $serializable = $callback;
```
shanerbaner82, line 88: **Every `$this`-bound fluent callback now pays a durable-tier tax per registration.** Pre-PR these early-returned. Now each user action pays: `ReflectionClosure` re-tokenizing the component's source file (its static cache only amortizes within one process — paid fresh per request on the stateless web target), `serialize()` of every captured `use` var, and a synchronous SQLite `Cache::put` — multiplied by each outcome callback on the builder (success + cancel + denied = 3 writes per tap). A closure capturing decoded image bytes means a multi-MB serialize + disk write per tap. And unserializable captures (PDO handles etc.) now build + `report()` a full exception per registration where the old code skipped quietly.

Cheap mitigations: size-cap the serialized payload and fall back to memory-tier-only with a debugLog; dedupe/downgrade the catch's `report()`; if action-request latency matters, defer the `Cache::put` batch to `App::terminating` — still completes before any OS kill since the native activity launches after the response.

src/Edge/NativeComponent.php
```diff
@@ -1922,8 +1922,24 @@ private function fireNativeCallback(string $eventName, array $payload): void
             return;
         }
 
-        if (is_string($callback) && class_exists($callback)) {
-            $callback = app($callback);
+        if (is_string($callback)) {
+            if (class_exists($callback)) {
```
shanerbaner82: **`class_exists()` is checked before `method_exists()` — and it's case-insensitive + autoloading, so a handler method whose name matches any loadable class is silently hijacked into `app()` resolution.**

Probe-verified: `class_exists('error')`, `'exception'`, `'datetime'`, `'generator'` are all true (PHP built-ins, lowercase), and Laravel facade aliases (`'log'`, `'view'`, `'auth'`, ...) collide once the alias has been materialized earlier in the process — making the behavior warm-up/order dependent. A component method named e.g. `error` registered via `->mediaSelected('error')` resolves `app('error')` instead; `call_user_func` on the non-invokable result throws after the `finally` has already consumed the callback.

The component's own method is unambiguously more specific: check `method_exists($this, $callback)` **first**.

src/Concerns/HandlesNativeCallbacks.php
```diff
@@ -43,6 +43,17 @@ public function on(string $eventClass, Closure|array|string $callback): static
...
 
+    /**
+     * Register a callback for the operation's SUCCESS event without naming
+     * it — `->onSuccess(fn ($event) => …)` reads the same on every builder
+     * (photoTaken, mediaSelected, videoRecorded, …). Honors a custom
+     * ->event(Custom::class) override, since that replaces $eventClass.
+     */
+    public function onSuccess(Closure|array|string $callback): static
+    {
+        return $this->on($this->eventClass, $callback);
```
shanerbaner82, line 59: **`onSuccess()` snapshots `$this->eventClass` at call time, so `->onSuccess($cb)->event(Custom::class)` registers under the wrong event and silently never fires** — while this docblock says the `->event()` override is honored, with no ordering requirement stated.

`event()` only reassigns `$eventClass` and the builder sends the FINAL value to native (`'event' => $this->eventClass`); nothing re-keys earlier registrations. Native emits `Custom`, `resolve()`/`resolveByEvent()` look up `Custom`, miss, and the callback is dead with its durable copy lingering until TTL. The named-method path at least throws `BadMethodCallException` in one order; `onSuccess` is silent in both.

Defer event resolution to dispatch time (register at `toPayload()`/send, not at `onSuccess()`), or document the ordering constraint.

shanerbaner82 (commented):
Review of `f7424a7`. The core stdClass-carrier trick is sound — verified empirically that private access, `self::`/`static::`, and scope all survive the serialize round-trip, and it's deliberately deeper than SerializableClosure's native bound-object support (which would snapshot the whole component). But the machinery around it has real holes: 10 verified findings, several **reproduced with live probes**. Six are inline comments; the four below touch code outside the diff so they're here. Findings 1–4 (the two below marked must-fix plus the two inline must-fixes) all break the PR's stated purpose — surviving process death — so they should land before merge.

---

**1. (Must-fix) `DispatchEventFromAppController` crashes on the new method-name shape and destroys the callback** — `src/Http/Controllers/DispatchEventFromAppController.php:74`.

The controller is a second consumer of `NativeCallbacks` this PR never touched. It only handles class-strings (`is_string && class_exists`), so `->mediaSelected('onPicked')` falls through to bare `call_user_func('onPicked', ...)` → uncaught TypeError, 500. Worse: in the cold-cache case, `NativeCallbacks::resolve()` has already `Cache::pull`'ed the durable copy and `forget()` is never reached — the callback is permanently lost before the Edge loop could fire it. Both shells POST result events through this controller whenever a WebView exists (`NativeActionCoordinator.kt:80`, `ContentView.swift:391`), so the method-name shape **cannot work at all in WebView-mode apps**, despite the PR comment advertising it as the shape that survives the stateless web target.

Fix: mirror the Edge logic — peek-and-bail on non-class strings (like the existing `$this`-closure guard at ~line 65) — and wrap invocation in try/finally.

---

**2. (Must-fix) The headline gallery-kill scenario silently fails: id-less events can't reach the durable tier** — `src/Support/NativeCallbacks.php:121`.

`resolveByEvent()` scans only the in-memory tier. When Android kills the app during a gallery pick, both the Kotlin-side `pendingGalleryId` and PHP's `static::$memory` die — exactly the case this PR's own `docs/native-callback-api-design.md` (Finding 5) documents: the gallery result arrives with **no id**. `fireNativeCallback` can't use `resolve($id, ...)` and `resolveByEvent()` iterates an empty array, so the durable copy just written sits in cache under a key nobody can reconstruct until the 5-minute TTL. Every new test passes an explicit id, so the gap is untested.

Fix: a register-time index key (e.g. `native_cb_latest:{eventClass}` → id), since the cache store is non-enumerable.

---

**3. Durable closures rebind to whatever component is active at fire time** — `src/Edge/NativeComponent.php:1949`.

After a real OS kill the app cold-starts at `/` — nothing restores the screen stack (`.hot_restart` is written only by the dev hot-reload path). The result event resolves ScreenA's durable closure and `\Closure::bind($callback, $this, static::class)` rebinds it to the home screen. Probe-verified: `$this->picked = ...` becomes a silent dynamic-property write (deprecation only), or throws into `renderErrorScreen` if a same-named incompatible typed prop exists.

Fix: persist the registering component class with the durable entry and skip-and-report when the live `$this` isn't an instance of it.

---

**4. Durable cache key has no session/user scoping** — `src/Support/NativeCallbacks.php:70`.

`native_cb:{id}:{eventClass}` — on the stateless multi-user web target the shared app-wide cache is the ONLY tier, and the documented fixed-id API (`->id('closure-pick')`, used in every test/example) collides across users: user B's registration overwrites user A's serialized closure, and whichever result event posts first `Cache::pull`s the survivor — user A's event can execute a closure carrying user B's captured `use` vars (correctness bug + data-leak vector), and A's own callback is gone. Auto-generated UUID ids are safe; the fixed-id API is the trap.

Fix: fold a session discriminator into `key()` when not on-device.

---

Cut under the 10-finding cap but worth a glance while in here: the `#[On]` key-collision widening (when both the bare and prefixed spellings coexist, the last-declared handler silently wins); the dropped function-name/`'Class::method'` string fall-through (undocumented, unused in repo — fine if intentional, worth a docblock line); `Str::start($event, 'native:')` instead of the hand-rolled ternary in `On.php` (also collapses an already-doubled prefix, which the ternary passes through); `[$this, $callback]` instead of the wrapper closure in `fireNativeCallback` (skips the redundant re-reflect/rebind); and the second durable test re-runs the first's whole flow for one extra assertion — could fold into one test.

PR #305, opened 2026-08-08
Headless build/verify commands, durable fatal capture, and report() for EDGE exceptions

shanerbaner82:
## What this is

Three things core should own for headless/agent-driven development. It deliberately does **not** ship an observation or devtools stack — see [Scope change](#scope-change) below.

## 1. Headless commands

New: `native:devices`, `native:screenshot`, `native:status`, `native:open-url`. Reworked: `native:tail` (iOS parity, `--lines`, `--follow`, `--json`) and `native:run --json`.

`--json` mode never prompts and never `exit()`s past its own output: deterministic device resolution (explicit `--device`, then a recorded `nativephp/agent-device.json` claim, then the single booted target), structured errors carrying a `stage` (`validate|devices|build|install|launch|verify`), a `logTail` on build/install failures, and an exit code that is 0 only when `ok:true`. Success is verified empirically — the app must actually be installed and running — which also catches unchecked `simctl` failures.

## 2. Durable boot/fatal capture

`bootstrap/shared/devtools-boot-report.php` — dependency-free by design, because it runs where Laravel is half-booted or already dead. Chained `set_exception_handler`, shutdown-time fatal reporting, and an `error_get_last()` re-entry check for the persistent runtime, where a fatal's Zend bailout is caught by the native host so shutdown functions never fire. `bootstrap/ios/native.php` also gains its previously missing top-level try/catch.

Laravel's `HandleExceptions` bootstrapper *replaces* rather than chains `set_exception_handler`, so the handler installed by the bootstraps is discarded during boot; `NativeServiceProvider` re-installs it afterwards, guarded against stacking.

This is the one failure class no observer or reporter can cover — there is no framework left to report through. Core writes the spool and never reads it back; draining it is a devtools package's job.

## 3. `report()` for EDGE exceptions

The EDGE loop catches around render, dispatch and polls and draws its own error screen — and told nobody. An EDGE screen could throw and Sentry, Flare, Telescope, Bugsnag and `laravel.log` would all see nothing. Native screens now report through Laravel like the rest of the app.

Reported once per throwable: `renderErrorScreen()` also serves the overlay repainting an exception already on screen, so it is guarded on identity. Polls are now caught like render and dispatch — a throwing `#[Poll]` previously unwound the whole session, losing the navigation stack.

## Scope change

Earlier revisions of this branch added an observation stack: an `ExceptionReporter`/`CrashRelay` seam, an `EdgeTicker` seam with a 250ms idle tick, `dumpTree()`/`dispatchSyntheticEvent()`/`callbackRegistry()`, and watch-mode listener plumbing. **All of that is withdrawn.**

[`stillat/mobile-air`](https://github.com/stillat/mobile-air) branch `instrumental-runtime` already defines `Native\Mobile\Edge\Contracts\RuntimeObserver` — a typed observer contract with `componentPublished`, `dispatchStarting`, `dispatchFinished` and `failed(RuntimeFailed)` — and [`stillat/tesseract-native-collector`](https://github.com/stillat/tesseract-native-collector) already consumes it, with EDGE element instrumentation, a command pump, telemetry forwarding and an MCP surface.

That design is better and it has a shipping consumer. `RuntimeFailed(ComponentContext, Throwable)` is what the `ExceptionReporter` seam was reaching for, typed and part of a coherent lifecycle API. And because an observer API is push, nothing needs a turn on the render loop: `EdgeTicker` existed only to service a polled command file, and the idle-tick defects that came with it — a 250ms cap that re-rendered an idle screen ~4×/second — are gone with the mechanism that caused them. `nextEventTimeout()` is back to `-1`.

The three pieces kept are the ones an observer cannot provide: build-and-launch, pre-runtime capture, and the reporting seam itself.

## Verification

893 passed, Pint clean. Unit-tested: device resolution, the bootstrap handlers (including a real OOM fatal and the re-install path), spool rotation, app-frame extraction, and report-once on the error overlay — each verified to fail when the fix under test is reverted.

Reviewed across four rounds, including two independent fresh-context passes.


shanerbaner82:
## Independent review (fresh-context session, adversarial pass over this PR + the nativephp/devtools plugin)

**Verdict: needs fixes before merge.** Five blockers:

1. **[major] EdgeTicker busy loop** — the 250ms idle cap makes every idle EDGE screen re-render+republish ~8x/sec (each timeout `continue`s into a full render; each publish re-dirties the snapshot so `dumpTree()` renders again + disk write). Fix: mark the snapshot dirty only on real tree change; don't force re-render on idle ticks.
2. **[major] Spool cursor protocol loses crashes** — the PHP writer advances `spool.cursor` past its own line on a successful POST, implicitly marking earlier *failed* lines delivered; the native drainer then never sends them. Concurrent writers can regress the cursor → duplicate sends (no id dedup at the listener). Fix: only drainers advance the cursor and/or dedup by event id at the listener.
3. **[major] Ticker gated on `app.debug` alone** — a release-signed build shipped with `APP_DEBUG=true` services `ui-command.json` and exports the screen tree. Require the `devtools.json` provisioning gate (like Transport) or a native debug-build signal.
4. **[major] Release builds spool unconditionally** — bootstrap fatal handlers always append to the device spool with no cap/rotation; recurring production fatals grow it unboundedly. Cap + gate the spool write.
5. **[major] Listener host reuse breaks physical-iOS streaming** — reuse only pid-checks `listener.json`; a loopback listener from a simulator session gets reused for a LAN-provisioned device → every POST fails silently. Compare bind host on reuse.

Notable minors (fast-follow): `--json` can still prompt on the `--build` select; drainer byte-offset math breaks on invalid UTF-8 (split on raw-byte `\n` before decoding); reused listener replays the entire historical event log into the console; recurring-fatal dedup hides crash loops; Android `open-url` arg reaches the device shell unquoted; `tap "Follow"` substring-matches "Following"; `listener.json` (bearer token) + `events.jsonl` never gitignored; slowloris + non-constant-time token compare on the LAN-bound listener path.

Clean areas verified: EventLog locking/seq, CrashRelay/LoopTick re-entrancy, MCP server (no path traversal), pull session-cutoff logic, JS capture double-gating, `--json` exit codes, version-guard degradation against older cores.

shanerbaner82:
## Agreed fix split (two review sessions, converged)

Two independent reviews (the comment above, plus a core-side review) converged on the same defects. Ownership agreed between the reviewing session and this branch's author. Recording it here so it survives both sessions.

**Core (this PR)** — assigned to the reviewing session, *not started, pending their user's go-ahead*:
1. **EdgeTicker busy loop** (highest priority). `nextEventTimeout()` capping at 250ms makes the idle branch `continue` into a full render+publish 4x/sec; the plugin's `dumpTree()` adds a second render per tick. Fix: `EdgeTicker::tick()` returns `bool` ("I mutated state, re-render me"); the idle branch loops back to `wait_event` instead of falling through, breaking out only on a real event, a fired poll, or `tick() === true`. **Hard coupling:** PHP has no void→bool covariance, so the plugin's matching one-line change must land with it — any debug build resolving the binding in between fatals.
2. `native:run --json` still prompts on the `--build` select (`RunCommand.php:174`) — default to `debug` in json mode.
3. `ManagesDevtoolsListener::startDevtoolsListener()` reuse branch: seed the events.jsonl read offset (else the watch console replays the whole log), and validate host *coverage* — `0.0.0.0` covers loopback+LAN, `127.0.0.1` covers loopback only; only the loopback→LAN upgrade kills the incumbent by recorded pid and respawns (with a visible note, since it can interrupt a concurrent `native:watch` on the same repo).
4. `app_frame` in `devtools-boot-report.php` (first trace frame under base_path, outside vendor/) — currently only the plugin's `Payload` sets it, so boot/shutdown fatals lack it.
5. Size cap on the spool append in `devtools-boot-report.php` — it writes unconditionally in every build type with no cap or rotation.
6. `rm` the pushed `/data/local/tmp/nativephp_devtools.json` after the `run-as cp` — it carries the bearer token and is world-readable.
7. Quote the URL for the *device* shell in `native:open-url` (adb re-parses argv on-device).
8. Re-install the uncaught-exception handler after `$kernel->bootstrap()` — Laravel's `HandleExceptions` replaces rather than chains, so `kind:exception, mode:uncaught` events never arrive today. Also clear the process-lifetime `__nativephp_devtools_reported` flag per cycle.

**Plugin (`nativephp/devtools`)** — **UNOWNED**, in priority order:
1. **Delivery protocol.** The defect is one line: `Transport::dispatch()` writes `spool.cursor` to the end of *its own* line on a 2xx, marking every earlier byte delivered — including lines whose POST just failed. Note this loses **core-written** events too: `devtools-boot-report.php` appends *and* does its own best-effort POST without recording the outcome, so a failed helper POST followed by a successful Transport POST strands the helper's line. Fix: PHP stops writing the cursor entirely; the native drainer owns advancement; the listener dedups on the event `id` (already a random 16-byte id) so redelivery is harmless. A cursor-blind second writer is *not* a problem — appends land at EOF, beyond any cursor.
2. Gate the ticker on `EndpointConfig::active()` (the provisioning file) in addition to `app.debug`, so a release build shipped with `APP_DEBUG=true` can't service `ui-command.json`.
3. `tick(): bool` signature change — must land with core item 1.
4. Listener: `hash_equals` for the token compare, and a per-connection read deadline (currently a slow client can hold the single-threaded accept loop).
5. `native:ui tap`: exact text match before substring (`tap "Follow"` currently matches "Following").
6. Drainer UTF-8 byte math (iOS + Android): split on the last raw-byte `\n` *before* decoding — the Kotlin lenient-decode inflation and the Swift strict-decode stall share that root cause.

Not owned by either: gitignoring `nativephp/devtools/` (listener.json holds the session token) belongs in the app skeleton / mobile-starter.

shanerbaner82:
## Implementation order + acceptance criteria

Companion to the fix split above. Recording the reviewing session's acceptance criteria here so whoever implements the 14 items has the full spec in one place.

**Order:**
1. **Core 1 + Plugin 3 together, as one unit** — `EdgeTicker::tick(): bool` + the loop restructure + the plugin's matching signature change. Push together; never leave a gap where a debug build resolves the binding against a mismatched signature (hard fatal — no void→bool covariance).
2. **Core 2** (`--build` defaults to `debug` in json mode) — smallest fix, unblocks reliable agent runs.
3. **Plugin 1** (delivery protocol) — highest-value plugin item.
4. **Core 3** (reuse branch: offset seed + host coverage + kill/respawn with a visible note).
5. **Core 8** (re-install handler after bootstrap; clear the process-lifetime reported flag per cycle).
6. Remaining in any order: Core 4, 5, 6, 7 and Plugin 2, 4, 5, 6.

**Core 1 must satisfy all four** — a naive dirty-check has three ways to silently break the UI:
- (a) Idle screen with a ticker bound performs **zero** renders — not fewer, zero.
- (b) After a `ui-command` invoke, the screen repaints on the **very next** loop iteration. A fix that stops the busy loop by dropping legitimate repaints is worse than the bug.
- (c) A poll that fires and mutates state still triggers a re-render. `runDuePolls()` returns void today; change it if it needs to report "something changed" — core-internal, no coupling.
- (d) With **no** ticker bound, behavior is byte-identical to pre-PR: `nextEventTimeout()` returns `-1` and the loop parks. Verify this path explicitly — it's the one every non-debug user is on.

**Plugin 1 acceptance:** a crash whose POST fails is still delivered by the drainer on the next drain, **and** redelivery produces no duplicate lines in `events.jsonl`. Both halves matter — id dedup is what makes the safe advance policy possible.

**Core 8 acceptance:** an uncaught throwable in a fully booted app actually produces a `kind:exception, mode:uncaught` event on device. If it still doesn't after the fix, stop and re-open the diagnosis rather than forcing it.

**Rules:** no merge, no push to main, no force-push; work on `feature/devtools-loop` and the plugin's own branch. If a fix turns out bigger than the spec implies, stop and re-scope rather than expanding silently.

**Evidence required on completion:** commit SHAs on both repos + confirmation both pushed; actual test and pint counts per repo (not "green"); for Core 1, a render/publish count over a fixed idle interval with a ticker bound, **before and after** — "I verified the behavior" is precisely what failed here, so a number is required; for Plugin 1, how the failed-POST-then-drain path was tested (and say so plainly if it couldn't be tested on a device); plus any deviations from spec, and anything not done, with reasons.

PR #307, opened 2026-08-09
Expanded Developer Tooling Hooks

JohnathonKoster:
We're cookin' with gas now 🔥

----

This PR adds a small set of opt-in extension points for packages that inspect, debug, automate, or test a running Native PHP Mobile application.

The goal is to let developer tools observe the real native runtime without replacing Native PHP's router, renderer, event loop, or testing APIs. Nothing is registered by default, so applications that do not use these extension points keep their existing behavior.

Although Tesseract is the first consumer, this PR contains no Tesseract transport, protocol, UI, commands, or other product-specific implementation.

### Why this is needed

A useful native debugger must be able to connect several pieces of information that currently stop at different runtime boundaries:

- which component instance is mounted or resumed;
- what interaction is being dispatched to that component;
- what state and timings resulted from the dispatch;
- which native tree was actually accepted by Android or iOS;
- where a rendered native node came from in the application's Blade source;
- how an installed package can handle its own commands safely; and
- how existing `Native::test()` and `Native::visit()` calls can report their progress.

These hooks share component identity and publication boundaries. Shipping only some of them would leave tools with an incomplete or ambiguous view of the application, so they are proposed together in one branch with separate logical commits for review.

#### Correlated screen lifecycle events

`ScreenMounted`, `ScreenResumed`, and `ScreenUnmounted` gain an optional process-local component ID populated by the router.

Class name and URI are not enough when the same screen can exist more than once in a navigation stack. The component ID lets a tool correlate lifecycle, state, and dispatch events with the exact live instance. Existing two-argument event construction remains valid.

#### Runtime observers

A new `RuntimeObserver` contract reports:

- completed component publications;
- interaction and native-dispatch start and finish boundaries;
- component context and process-local identity;
- render, serialization, publication, and dispatch timings; and
- failures caught by the runtime before normal error recovery.

Observers can be registered and unregistered independently. Observer failures are isolated from application behavior, and the empty registry avoids event snapshot work on hot paths.

#### Namespaced package commands

`NativeEventHandlers::register()` and `unregister()` allow an installed package to own commands such as `vendor:command` without adding methods to application components.

A handler explicitly claims an event with `Handled` or allows normal component dispatch to continue with `Pass`. Invalid, declined, or failed handlers fall through safely, and requiring a namespace avoids collisions with application and core events.

#### Element metadata hooks

Named pre-parse attribute transformers run consistently in both the object-backed element-tree and streaming render paths.

A package can retain source locations, stable keys, or original classes as private node props before Tailwind parsing removes or transforms them. Transformer failures are isolated and leave the original attributes available to the normal renderer.

#### Accepted native-tree observation and decoration

Matching Android and iOS registries allow packages to:

- observe the tree actually accepted by the platform renderer;
- identify publications with monotonically increasing IDs;
- replay the latest accepted publication to a late subscriber;
- unregister observers;
- decorate nodes through ordered, named decorators; and
- read generic node props for serialization.

The observation point is after parsing and reconciliation, at the existing accepted-publication boundary. A tool therefore sees the hierarchy the platform uses rather than an earlier PHP or decoder approximation.

Registries notify immutable subscriber snapshots outside their locks. On iOS, the normal concrete SwiftUI view path is retained unless a decorator is registered.

#### Extensible native test reporting

`TestableComponent::useHarness(?string $class)` allows a reporting package to extend the existing native test harness used by both `Native::test()` and `Native::visit()`.

The built-in harness remains the default, and invalid registrations fall back safely. Application tests continue to use the standard NativePHP API.

#### Android JNI lifecycle compatibility

Android's native bridge resolves `startWatching()` as a static JNI method, but the Kotlin companion method did not expose a Java-static entry point. Adding `@JvmStatic` makes that lookup succeed so the native input watcher starts and physical device events can be observed.

### Relationship to existing tree observation

This PR builds on the upstream observation model instead of creating a second source of truth:

- `TreeObservers` remains responsible for generic PHP tree, event, and navigation observation.
- `RuntimeObserver` adds component state, dispatch correlation, timings, and failures that cannot be reconstructed from a serialized tree.
- The Android and iOS registries report the post-reconciliation tree accepted by the platform renderer.
- Screen lifecycle events remain the source of mount, resume, and unmount transitions, now with the same component identity used by runtime events.

These layers describe different boundaries and do not publish duplicate definitions of navigation, lifecycle, or component state.

### Example tooling flow

When inspecting a counter interaction:

1. The accepted-tree observer reports the on-screen hierarchy and publication ID.
2. Captured props associate the selected native node with its Blade source and active classes.
3. A decorator highlights the node without replacing the normal renderer.
4. A namespaced command invokes an allowed component action.
5. Runtime observers report the dispatch, before/after state, timings, and resulting publication.
6. If navigation occurs, lifecycle events identify the exact component instance that mounted, resumed, or unmounted.

### Compatibility and safety

- All extension registries are empty by default.
- Existing rendering, navigation, dispatch, and testing APIs retain their defaults.
- Existing two-argument screen lifecycle event construction remains supported.
- No debugger transport, protocol, command implementation, or selection styling is added to core.
- Observer, transformer, handler, and decorator failures are isolated from normal application behavior.
- Captured metadata increases node payload size only for attributes a package explicitly registers.
- Runtime events expose live component instances for authoritative inspection; consumers must treat them as read-only and remain responsible for redacting application data.

### Native PHP dependency boundary

This branch is the only Native PHP branch required by the Tesseract native collector.

### Review notes

The changes are separated into focused commits for lifecycle correlation, runtime observation, package commands, element metadata, platform rendering, native test reporting, and the Android JNI bridge.


shanerbaner82 (commented):
Reviewed the branch and reproduced the baseline: **888 passing, Pint clean**.

Also merge-tested against #305, since we both touch `NativeComponent.php` and `NativeRouter.php` — **auto-merges with no conflicts, 911 passing together**. The two error-path hooks compose correctly: `reportToLaravel()` hands the throwable to Laravel's reporter, `RuntimeObservers::failed()` notifies tooling, and both independently guard against re-firing on an overlay repaint. Nice to see `lastNotifiedRuntimeFailure` as a dedicated field — that's the more robust of the two guards.

The "free when unused" discipline throughout is the thing I'd most want to keep: all three publish sites guarded by `RuntimeObservers::any()` before any allocation, and `dispatchNativeEvent()` going further with a complete early-out so the zero-observer path is byte-identical to before. Same for the transformer and capture sites in `NativeElementCollector`.

Findings below, ranked. Everything in the native files is reasoned from reading — **I could not compile or run the Kotlin/Swift**, so treat 1–4 as "worth measuring" rather than settled.

---

### 1. iOS pays a lock per node, per render — Android doesn't

`SwiftUINodeRenderer` calls `NativeNodeDecoratorRegistry.shared.currentPipeline` for every node, and `currentPipeline` takes `NSLock` on every read. That cost lands in every app, including release builds with zero decorators registered.

The Kotlin registry solves the identical problem lock-free:

```kotlin
@Volatile private var snapshot: List<NativeNodeDecorator> = emptyList()
fun isEmpty(): Boolean = snapshot.isEmpty()   // no lock on the read path
```

Swift could hold the pipeline somewhere readable without locking — `OSAllocatedUnfairLock`, or an atomic box — and keep the existing lock for writes only.

Worth noting the surrounding design is right: pre-building the pipeline instead of looping decorators per node, and returning `nil` when empty so the renderer keeps the concrete view and never wraps in `AnyView`. That last part matters a lot for SwiftUI diffing and is easy to get wrong.

### 2. `NativeTreeObserverRegistry.publish()` has the same asymmetry

Kotlin short-circuits on `@Volatile hasObservers` before touching the lock. Swift takes `NSLock` on every accepted publication. Cheaper than #1 (per frame rather than per node), same fix.

### 3. Both registries retain `latestPublication` for the process lifetime

It's assigned on every publish regardless of whether anyone is subscribed, so a release build permanently holds a reference to the last accepted tree.

That's the price of replay-to-late-subscriber, which is a genuinely good feature — but it's currently paid by apps that will never have a subscriber. Either gate the retention on `hasObservers` (losing replay for the very first subscriber) or document the trade.

### 4. A subscriber can receive the same publication twice

`register()` reads `latestPublication` and replays it outside the lock, while a concurrent `publish()` may already have that observer in its snapshot. Both deliver the same publication.

Well mitigated in practice, since `Publication.id` is monotonic and consumers can dedup on it — but it's currently something a tool author has to discover rather than something the contract states.

### 5. `NativeEventHandlers::register()` accepts `':'` and `native:…`

I probed the validation: `':'`, `'native:deeplink'` and `'__deeplink:x'` are all accepted.

Not exploitable today — core's raw event names come from `#[On]` values (class names, no colons), so the colon requirement does isolate the namespace in practice. But `native:` is core's own prefix, and I'd reserve it explicitly rather than depend on that continuing to hold.

### 6. Inconsistent validation posture between the two new registration APIs

`NativeEventHandlers::register()` throws on a bad event name. `TestableComponent::useHarness()` accepts any string and falls back silently at use time via `is_subclass_of`. The fallback itself is correct and safe — but a typo'd harness class simply never applies, with no signal to the tool author.

### 7. Small wording point in the description

> Transformer failures are isolated and leave the original attributes available

True for a *throwing* transformer — caught, `continue`, prior `$attrs` preserved. Not true for one that returns a valid-but-wrong array, which is accepted wholesale. Inherent to the extension point rather than a defect, but the sentence promises a bit more than the code does.

---

### Verified specifically

- Snapshot-taken-inside-lock, notify-outside-lock in both native registries — correct, and the thing that prevents reentrancy deadlock
- Observer failures isolated (PHP `notify()`, Kotlin `runCatching`). Swift has no equivalent, but the closure type is non-throwing, so that's a language difference rather than a gap
- Backward compatibility on the lifecycle events holds — `componentId` is a trailing optional param, two-arg construction still valid
- `NativeEventHandlers::dispatch()` runs ahead of the deeplink branch and `#[OnNative]` listeners, with `Handled` short-circuiting correctly
- Attribute transformers are wired into both the object-backed and streaming paths, and guarded so they cost nothing unregistered

The observer model is a better answer than the polled seams it replaces, and the accepted-tree boundary is the right place to observe from — post-reconciliation beats any PHP-side approximation. Main ask is bringing the iOS read paths up to the Android ones.

shanerbaner82 (commented):
Second pass on the branch as it stands now (17 commits). Full PHP suite: **918 passing, 3 skipped, 4 risky** — clean.

Last round's findings came back well. Addressed: **#1** (the pipeline is now read once in `NativeTreeRenderer.body` and handed down via `EnvironmentKey` instead of per-node, and `OSAllocatedUnfairLock` replaces `NSLock`), **#2** (`publish()` short-circuits on an `Atomic<Bool>` before touching the lock), **#5** (`register()` now rejects unscoped, `native:`-prefixed and `__`-prefixed names, with parametrised coverage), **#6** (`useHarness()` throws on non-subclass, abstract, private ctor, and over-wide constructors). **#3/#4** got doc comments rather than code changes, which is a fair resolution for both.

The same "free when unused" discipline holds throughout the new work, and I want to be clear up front that it's the reason the bulk of this is easy to say yes to.

---

## Verdict

I'd like to hold **one commit** — `82be218f "Defer iOS tab reconciliation"` — and merge the other sixteen.

That commit is 32 lines across three Swift files and reverts cleanly. Everything else in the PR — roughly 1500 of the 1538 added lines — is either inert by default or purely additive, and I have no reservations about it.

## Why the surface is smaller than the diff

Almost none of this can reach an app that hasn't opted in. I checked each seam:

- `RuntimeObservers` — every call site guarded by `any()`, and `dispatchNativeEvent()` goes further with a complete early-out, so the zero-observer path is byte-identical to before.
- `NativeEventHandlers` — a single `isset()` on native events only.
- Attribute transformers and captured attributes — guarded on empty registry; the `createElement` refactor only executes once a package registers a capture.
- Tree observer registries on both platforms — `hasObservers` short-circuit ahead of any lock.
- `TestableComponent::useHarness` — test path only.
- `@JvmStatic` — strictly a fix, and `NativeElementBridgeJniAbiTest` is a genuinely good regression guard. It catches the whole class of silent JNI lookup failure rather than just `startWatching`.

The risk in this PR is concentrated in about 36 lines of iOS and Android runtime changes that aren't developer-tooling hooks and happen to share the branch.

## The two I'd like to resolve before merge

**1. Tab layouts now render one runloop turn stale — `NativeRootTabsRenderer.swift:92`**

This is the one I'd most like eyes on, because it isn't a race — it's deterministic. Removing `coord.cache(uri:node:)` leaves the deferred `receive()` as the only writer of `rootNodeCache`, and `PerTabContent.levelView` renders solely from that cache with a `Color.clear` fallback (`NativeRootTabsRenderer.swift:434`). So every publish under a tab layout paints the previous frame's node for one main-loop turn, and each tab's first publish paints `Color.clear`.

The removal is directionally right and I don't want to argue it away — `rootNodeCache` *is* `@Published`, so the old synchronous write was a real publishing-during-view-update violation and the doc comment claiming it was safe was simply wrong. But whether the deferral reads as imperceptible or as a flicker on every tab interaction is empirical, and I can't answer it from the diff. Could you grab a device capture of tab push/pop/switch on a tab layout?

**2. `NavigationCoordinator.reset()` replaces the `@Published` storage — `NavigationCoordinator.swift:191`**

```swift
_path = Published(initialValue: [])
```

Combine installs the `objectWillChange` publisher into each `@Published`'s internal storage once, and `path` is the only `@Published` on this class. Replacing the wrapper discards the installed publisher, so a subscription established before `reset()` is orphaned and `path` mutations stop emitting.

Working it through, I think it heals: `reset()` at `NativeElementBridge.swift:521` is immediately followed by `publishTree(...)`, which re-renders the parent, re-creates `NativeRootStackRenderer`, and re-subscribes its `@ObservedObject` — reinstalling a fresh publisher. That's very likely why it tests clean, and I'm not claiming it's broken today.

My concern is the shape of the dependency rather than the current behaviour. It rests on undocumented Combine install-once semantics *and* on SwiftUI re-creating that view struct on the next publish, and if either stops holding the symptom is a navigation stack that silently stops responding — no crash, no warning, nothing in the log. On the shared navigation singleton I'd rather not carry that, when `DispatchQueue.main.async { path = [] }` buys the same teardown quiet with none of the coupling.

Related: `tests/Unit/IosTabNavigationRendererTest.php:26` asserts the `_path` form is *present*, so adopting the alternative would require editing the test to permit it.

## Follow-ups — not merge gates

None of these change behaviour for an app without tooling installed. Happy for them to land separately.

**3. Decorator registries aren't observable — `SwiftUINodeRenderer.swift:45`, `NodeView.kt:294`.** Both read an unlocked/`@Volatile` snapshot from inside `body`/composition, and neither `register()` nor `unregister()` invalidates anything, so a decorator only takes effect on the next PHP publish. For a debugger highlighting the tapped node, the highlight won't appear until the app happens to re-render. Android's `snapshot` wants to be a `mutableStateOf`; iOS wants the pipeline pushed through a published property.

**4. iOS decorator toggle resets view identity — `SwiftUINodeRenderer.swift:203`.** `body` became `@ViewBuilder` + `if let decorate { … } else { rendered }`, so it's now `_ConditionalContent`. The `nil` path still avoids `AnyView` — the thing I flagged last time holds. But flipping the pipeline from `nil` to non-`nil` switches branches for every `NodeView` at once, and SwiftUI treats that as new identity: scroll offsets, in-flight animations and `TextField` focus all reset the moment a decorator registers. Combined with #3, enabling a highlight is a full state wipe. A modifier that no-ops on `nil` keeps identity stable across the toggle. Only affects tool users, so it's a Tesseract-side concern more than an apps concern.

**5. Android reload signals are adopted as already-seen — `MainActivity.kt:848,854`.** A signal written while the process was dead is now permanently ignored rather than firing once on next boot. Defensible for `.hot_restart`, since PHP consumes the file in the `Route::native` macro anyway, and the race described is real. Worth calling out in the description though — it's a hot-reload semantics change for every developer, in a PR that says no default behaviour changes.

**6. `publishCount` is overloaded — `NativeComponent.php:2175,2359`.** It was a memo-frame counter read only for the `% NPHP_FORCE_FULL_FRAME_EVERY === 1` heartbeat; it now also advances on the streaming path to feed `ComponentContext::$renderCount`. A component mixing paths (streaming render plus `publishPlaceholder()` / `renderErrorScreen()`, which go through `memoizedToArray`) shifts its heartbeat phase. Harmless today since the hash store is only consulted on the non-streaming path, but it couples an observability counter to a correctness mechanism. A separate `$renderCount` field costs nothing.

**7. `lastNotifiedRuntimeFailure` pins the Throwable — `NativeComponent.php:139`.** The dedup guard holds a strong reference to the last exception, transitively retaining its stack trace and every object bound in those frames for the component's lifetime. `spl_object_id($e)` or a `WeakReference` dedups identically without the retention. (Still think this guard is the more robust of the two error-path mechanisms — just the storage.)

**8. Captured-prop precedence differs by element path — `NativeElementCollector.php:422/473` and `523/569`.** Builtins use `$capturedProps + …` (captured wins on collision); plugin elements use `array_merge($capturedProps, $props ?? [])` (resolved wins). Same registry, same attribute, opposite outcome depending on element type.

**9. `Dispatch::$method` reports names that don't exist — `NativeComponent.php:1779`.** `__deeplink` is labelled `'__navigate'`, which isn't a method on the component; and when `NativeEventHandlers` claims the event or a fluent `->on()` closure handles it, `method` is `null` even though something ran. Three cases where a tool reading it as "the method about to be invoked" gets a wrong answer. Either document it as a best-effort label or add a `DispatchKind` case for handler-claimed events.

**10. `stopTransformingAttributes(null)` clears every package's transformers — `NativeElementCollector.php:149`.** Two packages register, one calls the no-arg form in teardown, the other silently stops working. `transformAttributes` takes a name and the targeted removal takes a name — it's the clear-all default on the same method that's the footgun.

**Nits.** `NativeEventHandlers::register()` validates case-insensitively via `/i` but `dispatch()` matches the key case-sensitively, so `Vendor:Cmd` registers and never fires. `NativeUIBridge.currentTree` is still directly settable (`NativeElementBridge.swift:1058` does so for teardown) — `private(set)` would make it impossible to bypass observation by accident. And `it('does not reuse component identities after an instance is destroyed')` passes trivially because ids come off a monotonic sequence; it never actually exercises WeakMap release.

## One process note

The riskiest changes here are the only ones with no runtime coverage — and can't have any, since the PHP suite can't exercise SwiftUI. The two tests added for them read Swift source with `file_get_contents` and match on regex, which pins the current text without saying anything about behaviour. I'd rather see a device capture carry that weight and drop those two tests, than have the assertions read as coverage they aren't.

---

To be clear about proportion: the tooling seams are the part that matters and they're in good shape. Post-reconciliation is the right observation boundary, the observer model beats the polled seams it replaces, and the zero-cost-when-unused discipline is held consistently across five separate extension points, which is not easy. My only real ask is that `82be218f` travel separately with a device capture behind it.

PR #313, opened 2026-08-11
[Masterclass] Fix items-start/items-stretch inversion on iOS and max-w-* being dropped on both platforms

### LINKED ISSUE #309: [Masterclass] iOS: items-start and items-stretch behave inverted on <native:column> compared to Android

Building a chat screen where each message bubble is only as wide as its own text and aligned to one side - incoming messages to the leading edge, outgoing to the trailing edge.

`items-start` is the class for that, and it works correctly on Android. The same markup on iOS produces the opposite result - the bubble fills the width.

### What happened?

Expected: `items-start` makes children hug their content and align to the leading edge; `items-stretch` makes children fill the parent's width. This is what Android does.

Actual on iOS: the two are effectively swapped. `items-start` stretches children to the full available width, and `items-stretch` makes them hug their content.

Image

So the same markup produces opposite layouts on the two platforms.

### How to reproduce the bug

Background colours make the child bounds visible: the pale blue wrapper shows the full available width, the inner block shows what the child actually occupies.

```blade
<native:column class="w-full p-4 gap-4">
    <native:column class="w-full items-start bg-blue-200 p-2">
        <native:column class="bg-red-300 p-3">
            <native:text>items-start</native:text>
        </native:column>
    </native:column>

    <native:column class="w-full items-stretch bg-blue-200 p-2">
        <native:column class="bg-green-300 p-3">
            <native:text>items-stretch</native:text>
        </native:column>
    </native:column>
</native:column>
```

Run on both platforms and compare.

- Android: the `items-start` child hugs its text, leaving blue visible to its right; the `items-stretch` child fills the width. Correct.
- iOS: the `items-start` child fills the width; the `items-stretch` child hugs its text. Inverted.

### Debug Output

```

   INFO  NativePHP Mobile.  

+--------------------+--------+
| Package Version    | 4.1.0  |
| PHP Version (Host) | 8.4.18 |
| OS                 | Darwin |
| OS Version         | 25.5.0 |
| Embedded PHP       | 8.4.24 |
+--------------------+--------+

   INFO  Installed Plugins.  

+--------------------------+---------+
| Package                  | Version |
+--------------------------+---------+
| nativephp/mobile-browser | 1.0.2   |
| nativephp/mobile-ui      | 0.3.0   |
+--------------------------+---------+

   INFO  Development Tools.  

```

macOS

iOS (Simulator), Android (Emulator)

### Notes

Android appears to have the correct behaviour, so this looks like an iOS-side
mapping issue rather than an ambiguity in what the attributes should mean.

### LINKED ISSUE #310: [Masterclass] max-w-* is ignored on all elements (both platforms)

Constraining a chat bubble's width so long messages wrap at a maximum instead of running the full width of the screen - the standard `max-w-*` case.

### What happened?

Expected: the element grows with its content up to the maximum, then wraps.

Actual: `max-w-*` has no effect at all. The element renders exactly as it does with no width constraint. This is the same on iOS and Android.

A fixed `w-[280px]` on the same element clamps correctly, so widths are parsed and applied — the max-width constraint specifically is not.

Not a case of the maximum being applied as a fixed width either: an element with `max-w-[280px]` and short content still hugs its content rather than padding out to 280px.

### How to reproduce the bug

Each block wraps a red test element in a pale blue parent, so the blue shows the full available width and the red shows what the child occupies.

```blade
<native:column class="w-full p-4 gap-5">

    <native:column class="w-full gap-1">
        <native:text>1. no constraint (baseline)</native:text>
        <native:column class="w-full items-start bg-blue-200 p-1">
            <native:column class="bg-red-300 px-3 py-2">
                <native:text>This is a longer line of text that will run as wide as it is allowed to run before wrapping.</native:text>
            </native:column>
        </native:column>
    </native:column>

    <native:column class="w-full gap-1">
        <native:text>2. w-[280px]</native:text>
        <native:column class="w-full items-start bg-blue-200 p-1">
            <native:column class="w-[280px] bg-red-300 px-3 py-2">
                <native:text>This is a longer line of text that will run as wide as it is allowed to run before wrapping.</native:text>
            </native:column>
        </native:column>
    </native:column>

    <native:column class="w-full gap-1">
        <native:text>3. max-w-[280px]</native:text>
        <native:column class="w-full items-start bg-blue-200 p-1">
            <native:column class="max-w-[280px] bg-red-300 px-3 py-2">
                <native:text>This is a longer line of text that will run as wide as it is allowed to run before wrapping.</native:text>
            </native:column>
        </native:column>
    </native:column>

</native:column>
```

Identical results on iOS and Android:

- Block 2 clamps correctly — red stops at 280px, blue visible to its right.
- Block 3 renders identically to block 1 — red fills the full width, as if the constraint were absent.

Also confirmed, with the same result: `max-w-sm` (named scale), `max-w-*` on `<native:row>` and on `<native:text>`, and `max-w-*` with short content (which stays hugged, so the maximum is not being applied as a fixed width).

### Debug Output

```

   INFO  NativePHP Mobile.  

+--------------------+--------+
| Package Version    | 4.1.0  |
| PHP Version (Host) | 8.4.18 |
| OS                 | Darwin |
| OS Version         | 25.5.0 |
| Embedded PHP       | 8.4.24 |
+--------------------+--------+

   INFO  Installed Plugins.  

+--------------------------+---------+
| Package                  | Version |
+--------------------------+---------+
| nativephp/mobile-browser | 1.0.2   |
| nativephp/mobile-ui      | 0.3.0   |
+--------------------------+---------+

   INFO  Development Tools.  

```

macOS

iOS (Simulator), Android (Emulator)

### Notes

`max-w-*` appears to be dropped rather than mismapped, and it is dropped consistently: both the named scale and arbitrary values, on column, row and text, on both platforms.

shanerbaner82:
Fixes #309, fixes #310. cc @shrutibalasawebdev

Both are in the EDGE layout path. Demos for each live in the super-native demo app under a new **Masterclass Fixes** group (`/masterclass/align-items`, `/masterclass/max-width`) — every section is a visual assertion with the expected result written next to it.

---

## #309 — `items-start` / `items-stretch` inverted on iOS

iOS had the two cross-axis branches transposed in `FlexContainer.placeSubviews`. The `stretch` case measured the child at its **natural** cross size (that's `start` behaviour), while the `start` case reused `childCrosses[i]` — the Phase 3 measurement taken against `crossAvail`, i.e. the full container cross (that's `stretch`). Android's `ComposeFlexLayout` was already right. The transposed branch even carried a comment saying "align to start (like Android)".

### Why the obvious swap isn't the fix

PHP omits `align_items` entirely unless an `items-*` class is authored, so it arrives as `0` — **the same value as `items-start`**. Nothing downstream can tell "the author wrote `items-start`" apart from "the author wrote nothing".

So swapping the branches also flips every *unclassed* container on iOS from fill to hug. I tried it; it emptied the demo app.

### What this PR does instead

Explicit start gets its own wire value. `0` now means UNSET and is deliberately **not** an enum case; `Start` moves to `4` on both `AlignItems` and `AlignSelf`:

| wire | meaning | iOS | Android |
|------|---------|-----|---------|
| `0` | unset — no `items-*` class | fills *(unchanged)* | content-sizes *(unchanged)* |
| `3` | `items-stretch` | fills — **was hugging** | fills *(unchanged)* |
| `4` | `items-start` | hugs — **was filling** | content-sizes *(unchanged)* |

Nothing unclassed moves, which is the whole point. On iOS, `0` and `3` now share the `default:` branch so unset behaves exactly as it always has. **Android needed no behavioural change at all** — `4` falls through the same `else` branches `0` always did; only the constant was added so it can't land somewhere wrong.

### Bonus: `self-start` was a silent no-op

Both renderers resolve `alignSelf > 0 ? alignSelf : align`, so while `Start` was `0` an authored `self-start` was indistinguishable from no class at all — on **both** platforms. Moving it to `4` fixes that for free.

### Deliberately not addressed

iOS fills on UNSET; Android content-sizes. That's a real parity bug in the same family, but closing it moves every unclassed container on one platform, so it wants its own decision and a device pass. It's called out as the one ⚠️ section in the demo.

> Note: `4` was previously `Align::BASELINE` before that was removed. I grepped both renderers — nothing reads `4` as baseline any more, so the reuse is safe.

---

## #310 — `max-w-*` ignored on all elements

Dropped in **four** separate places, which is why it read as a total no-op rather than a mismapping:

1. `TailwindParser` had no `min-`/`max-` branch **and** no arbitrary-value prefix — so `max-w-sm` *and* `max-w-[280px]` both parsed to `null`
2. `NativeElementCollector` never mapped the attrs onto the layout array
3. iOS `resolvedMaxWidth` only consulted `maxWidth` when *no* width mode was set, so `w-full max-w-*` ignored the bound
4. Android read min/max off the wire into `NativeUINode` and then never applied them to a modifier

The packed node **already carried min/max at offsets 78–93**, so no wire format bump was needed — only the plumbing at each end. `min-w-*`, `min-h-*` and `max-h-*` were missing for the same reason and are fixed by the same change.

On Android the modifier **order** is the behaviour: `widthIn` has to narrow the constraints *before* `fillMaxWidth` consumes them, or `w-full max-w-[280px]` fills the parent and gets clamped after the fact.

`@shrutibalasawebdev` explicitly called out that this shouldn't become a fixed width — the demo has a section for that (short content under `max-w-[280px]` must still hug).

### One deliberate gap

`max-w-full` / `max-w-screen` are left **unparsed**. Min/max ride the wire as bare floats with no companion size mode, so "100% of the parent" has nowhere to go. Leaving them unparsed surfaces them in the dropped-class diagnostics rather than looking supported-but-broken. Use `w-full` for that.

---

## Testing

- Full suite green: **881 passed**, Pint clean
- New test asserts `items-start` and a class-less column produce **different** wire output (`align_items: 4` vs the key being absent) — that's the invariant whose loss reintroduces exactly this bug
- New test proves `w-full max-w-[280px] min-h-16` reaches the wire as `max_width: 280.0` / `min_height: 64.0` / `width: 'fill'`
- Wire values for both alignment enums are re-pinned, plus an assertion that `0` resolves to no case so the "unset" slot can't leak back

⚠️ **The Swift and Kotlin changes are compile-unverified** — no iOS or Android build has been run against this yet.

### DIFF
--- a/resources/androidstudio/app/src/main/java/com/nativephp/mobile/ui/nativerender/ComposeFlexLayout.kt
+++ b/resources/androidstudio/app/src/main/java/com/nativephp/mobile/ui/nativerender/ComposeFlexLayout.kt
@@ -10,12 +10,15 @@ import androidx.compose.foundation.layout.RowScope
 import androidx.compose.foundation.layout.fillMaxHeight
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.heightIn
 import androidx.compose.foundation.layout.offset
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.width
+import androidx.compose.foundation.layout.widthIn
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.dp
 
 // MARK: - Flex Enums (match PHP/binary protocol values)
@@ -35,10 +38,25 @@ object JustifyContent {
 }
 
 object AlignItems {
-    const val START = 0
+    /**
+     * No `items-*` / `self-*` class was authored — PHP omits `align_items`
+     * entirely, so it arrives as 0 and each renderer applies its own default.
+     * On Android that default is content-sizing (this file never adds
+     * `fillMaxWidth()` for it), which is unchanged by #309.
+     */
+    const val UNSET = 0
     const val CENTER = 1
     const val END = 2
     const val STRETCH = 3
+
+    /**
+     * Explicitly authored `items-start` / `self-start`. Deliberately NOT 0:
+     * while Android renders START and UNSET identically, iOS does not, and a
+     * shared wire value meant iOS could not fix `items-start` without also
+     * flipping every unclassed container (mobile-air #309). Android behaviour
+     * is untouched — 4 falls through the same `else` branches 0 always did.
+     */
+    const val START = 4
 }
 
 object PositionType {
@@ -159,7 +177,7 @@ private fun FlexColumn(
     val alignment = when (align) {
         AlignItems.CENTER -> Alignment.CenterHorizontally
         AlignItems.END -> Alignment.End
-        else -> Alignment.Start // START and STRETCH both start-align; STRETCH handled per-child
+        else -> Alignment.Start // UNSET, START and STRETCH all start-align; STRETCH fills per-child
     }
 
     Column(
@@ -237,6 +255,38 @@ private fun FlexRow(
 }
 
 /**
+ * Apply `min_width` / `max_width` / `min_height` / `max_height` from the packed
+ * node as Compose constraint bounds.
+ *
+ * A bound of 0 is "unset" on the wire, so it maps to [Dp.Unspecified] rather
+ * than to a literal 0.dp — `widthIn(max = 0.dp)` would collapse the node.
+ *
+ * Shared by [buildChildModifier] and `NodeView`'s no-parent fallback so the
+ * constraints apply whether or not the node sits inside a flex container.
+ */
+fun Modifier.applySizeConstraints(layout: NodeLayout?): Modifier {
+    if (layout == null) return this
+
+    var mod = this
+
+    if (layout.minWidth > 0f || layout.maxWidth > 0f) {
+        mod = mod.widthIn(
+            min = if (layout.minWidth > 0f) layout.minWidth.dp else Dp.Unspecified,
+            max = if (layout.maxWidth > 0f) layout.maxWidth.dp else Dp.Unspecified
+        )
+    }
+
+    if (layout.minHeight > 0f || layout.maxHeight > 0f) {
+        mod = mod.heightIn(
+            min = if (layout.minHeight > 0f) layout.minHeight.dp else Dp.Unspecified,
+            max = if (layout.maxHeight > 0f) layout.maxHeight.dp else Dp.Unspecified
+        )
+    }
+
+    return mod
+}
+
+/**
  * Build per-child modifier for flex properties (weight, fill, fixed size, margin).
  */
 @Composable
@@ -273,6 +323,17 @@ private fun buildChildModifier(
             mod = mod.offset(x = negX.dp, y = negY.dp)
         }
     }
+
+    // Min/max size constraints (`max-w-*`, `min-h-*`, …).
+    //
+    // Applied BEFORE the fill/fixed/weight modifiers below so that a combo
+    // like `w-full max-w-[280px]` fills *within* the 280dp bound instead of
+    // filling the parent and being clamped after the fact. Compose narrows
+    // constraints outer→inner, so ordering here is the whole behaviour.
+    //
+    // 0 means "unset" on the wire (the packed node has no companion size
+    // mode for min/max), which is why each bound is gated on `> 0f`.
+    mod = mod.applySizeConstraints(layout)
 
     // Main axis: flex_grow or fill → weight
     val flexGrow = layout?.flexGrow ?: 0f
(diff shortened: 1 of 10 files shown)


shanerbaner82:
Added a second Android commit (`eaaba78`) while device-testing the demo screen.

`buildChildModifier` consulted `align_self` only to decide whether the child **filled** the cross axis, never where it **sat**. The container's Column/Row alignment parameter applies to every child uniformly, so an individual child can only move via `Modifier.align()` inside the layout scope — which was never called.

So `self-center` and `self-end` have never positioned anything on Android: a `self-end` child correctly hugged its content and then stayed on the leading edge. iOS honours `align_self` in FlexContainer's placement switch, so this was a straight platform divergence.

Pre-existing, but invisible until `self-start` became meaningful in the first commit — a screen exercising all three `self-*` values is what surfaced it. STRETCH is skipped (no placement needed; it's expressed as fillMaxWidth/Height) and 0/unset falls through so the child keeps inheriting the container's align-items.

Demo section 5 now covers both axes — a column for horizontal self-\*, and a row where the cross axis is vertical.

PR #319, opened 2026-08-11
Give the splash screen plugins a named patch marker to anchor to

unlocdavid:
Follow-up to #159, which was closed with:

> We're going to leave this functionality to Plugins rather than adding more complexity to the core. If there's some aspect of core that would make it easier for you to build a plugin for this feature, we would welcome it.

The splash behavior is a plugin now: [unlocnl/nativephp-enhanced-splash](https://github.com/unlocnl/nativephp-enhanced-splash). Building it turned up three things in core that made it harder than it had to be. This is the first, and the only one that affects every splash plugin rather than just that one.

No behavior change, no config, no dependency — comments and three path accessors.

## The contract already exists

`MainActivity.kt` currently carries this in `onFirstContent()`:

```kotlin
// The two lines below keep their original 12-space indent: splash-screen
// plugins (s2br/nativephp-mobile-splashscreen) patch them via exact-string
// match including that indentation. Do not re-indent.
```

Core is already maintaining a whitespace contract for a third-party plugin. There is now more than one splash plugin doing this, and between them they anchor on four sites in this file — an insertion point above `super.onCreate()`,
`visible = showSplash,`, `showSplash = false`, and `.background(Color.Black),`. All four are incidental details of the source that no release note would ever mention changing.

## What this does

Replaces the convention with named markers at the same four sites:

- `nativephp:splash-install` — where `installSplashScreen()` goes, above `super.onCreate()`
- `nativephp:splash-visibility` — the overlay's `visible` argument
- `nativephp:splash-release` — where `showSplash` is cleared on first content
- `nativephp:splash-backdrop` — the overlay's backdrop color

The contract is stated once, next to `showSplash`: each marker, the statement beneath it, and that statement's indentation are stable for the major version. That is cheap to keep, survives reformatting, and covers cases core never
anticipated.

The existing 12-space indentation is untouched, so the plugin the old comment was written for keeps working.

## Why markers rather than something better

The root cause is that the native project is copied from `resources/androidstudio` only at `native:install`, never per build, so a plugin cannot rely on starting from core's own source — it has to undo its own previous patch first, by exact
string. Re-emitting template-derived files each build would remove the need to patch at all, but it would also discard hand edits to the native project, which seems like the wrong trade. Markers are the cheap version.

## Also included

`NativePluginHookCommand` gains `coreTemplatePath()`, `iosTemplatePath()` and `androidTemplatePath()`.

A plugin that replaces a file core installs verbatim — `LaunchScreen.storyboard`, `SplashView.swift` — restores it by copying core's own copy back, so a core upgrade is never silently undone. Today that means hardcoding
`base_path('vendor/nativephp/mobile/resources/xcode/NativePHP/…')` into the plugin, which breaks under a custom vendor dir or a path repository. The accessors resolve it from the package's own location instead.

## Testing

`tests/Unit/Plugins/HookCommandTemplatePathTest.php` covers the accessors. The Kotlin change is comments only.

### DIFF
--- a/resources/androidstudio/app/src/main/java/com/nativephp/mobile/ui/MainActivity.kt
+++ b/resources/androidstudio/app/src/main/java/com/nativephp/mobile/ui/MainActivity.kt
@@ -95,6 +95,11 @@ class MainActivity : FragmentActivity(), WebViewProvider {
     // Last appearance pushed to PHP, so onConfigurationChanged (which also fires
     // on rotation) only emits AppearanceChanged when the theme actually flips.
     private var lastAppearance: String? = null
+    // Splash-screen plugins replace core's splash with their own by patching
+    // this file, which they can only do by matching its source exactly. The
+    // four `nativephp:splash-*` markers below name the sites they patch, and
+    // are a stable contract for this major version: each marker, the statement
+    // beneath it, and that statement's indentation.
     private var showSplash by mutableStateOf(true)
     // Gates composition of the heavy MainScreen tree (Scaffold + WebView)
     // until the runtime is booted and the WebView is ready. Until then the first
@@ -121,6 +126,7 @@ class MainActivity : FragmentActivity(), WebViewProvider {
     }
 
     override fun onCreate(savedInstanceState: Bundle?) {
+        // nativephp:splash-install — installSplashScreen() goes here, above super
         super.onCreate(savedInstanceState)
         instance = this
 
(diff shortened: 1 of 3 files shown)


simonhamp:
I know we've done this in some places, but patching a code file feels like it's a path to danger. I think you should probably come up with permanent code hooks in here that allow your plugin to work as intended

PR #340, opened 2026-08-15
Add mount dependency injection

shanerbaner82:
## What changed

- add a Native component lifecycle dispatcher that hydrates matching public route properties before calling `mount()`
- resolve Eloquent/UrlRoutable models and backed enums from route parameters
- inject ordinary class dependencies through Laravel's container
- update router, standalone, nested-component, and testing mount call sites
- add regression coverage for model binding, public-property binding, primitive parameters, container services, backed enums, and missing models

## Why

`NativeComponent` declared a parameterless `mount(): void`, so PHP rejected component hooks such as `mount(Click $click)` as incompatible overrides. Native components also received raw route values rather than Livewire-style implicit bindings.

The base class now exposes an internal `mountComponent()` dispatcher and deliberately leaves `mount()` undeclared, allowing application components to use Livewire-compatible hook signatures.

## Impact

Native routes such as `/counter/{click}` can now either hydrate `public Click $click` directly or inject the model through `mount(Click $click)`. Components may combine route-bound values with services resolved from Laravel's container.

## Validation

- focused dependency-injection tests: 4 tests, 26 assertions
- full Pest suite passed with 3,039 assertions (existing PHP 8.5 deprecation/risky notices remain)
- PHPStan: no errors in the affected framework classes
- Pint: clean
- validated end-to-end with a real Eloquent `Click` model in the Native scratchpad app

### DIFF
--- a/src/Edge/ImplicitlyBoundMethod.php
+++ b/src/Edge/ImplicitlyBoundMethod.php
@@ -1,0 +1,164 @@
+<?php
+
+namespace Native\Mobile\Edge;
+
+use Illuminate\Container\BoundMethod;
+use Illuminate\Contracts\Routing\UrlRoutable as ImplicitlyBindable;
+use Illuminate\Database\Eloquent\ModelNotFoundException;
+use ReflectionClass;
+use ReflectionNamedType;
+
+/**
+ * Resolves component method parameters with the same rules as Livewire:
+ * ordinary classes come from the container; routable models and backed enums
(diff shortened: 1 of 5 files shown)


simonhamp (approved):
Ship it

shanerbaner82:
Superseded by #343 (same mount DI, via `ComponentMethodInvoker` instead of `ImplicitlyBoundMethod`). Closing so we don't land two implementations.

PR #343, opened 2026-08-16
Add native component lifecycle foundations

shanerbaner82:
## Summary

Component actions, route binding, state updates, and component events now share explicit invocation and lifecycle paths. Components can receive container dependencies and route-bound models in `mount()` and actions, use ordered property-update hooks for dotted state paths, and dispatch bubbling or targeted component events.

The route matcher uses Laravel routes for constraints, prefixes, optional parameters, custom keys, scoped bindings, and soft-deleted models. Routability checks do not execute model bindings. Navigation binding failures stay in the current screen's error lifecycle; failed bindings during hot-reload stack restoration skip the invalid entry and continue.

## Behavior and compatibility

- Template actions must be public, non-static user methods. Framework internals and lifecycle hooks cannot be invoked as ordinary actions. Existing direct navigation helpers remain supported.
- Remove `parent::mount()` calls: the base declaration is removed so application `mount()` signatures can accept dependencies and route parameters.
- The former `protected dispatch(array $event)` runtime method is now `dispatchUiEvent()`. The new `public dispatch(string $event, mixed ...$params): ComponentEvent` API delivers component events. Update overrides or direct calls to the old runtime method.
- Rename ordinary template callbacks using reserved lifecycle names or prefixes (`mount`, `boot`, `updating*`, `updated*`, `hydrate*`, `dehydrate*`, etc.). Explicit `#[On]` listeners may use those names.
- The shared state pipeline retains `#[Locked]` protection and adds ordered `updating*` / `updated*` hooks for nested properties.
- `native:model` resolves view data from nested Blade partials.

This does not add validation, `#[Renderless]`, or `skipRender()`.

## Hardening

- Cleared controls bind an empty string to null for nullable model/enum parameters.
- Device and test-harness interaction paths share the same invocation rules.
- Throwing polls stop repainting the error overlay on every tick.
- Event assertion predicates inspect all matching dispatches. Targeted assertions require the target and payload to match the same event.
- Updated against `main` at `649b558`, retaining query-string route handling and current runtime changes.

## Attribution

Livewire-derived portions retain their source headers and Caleb Porzio attribution. The notice is in `LICENSE.md`, alongside the MIT terms.

## Verification

At `3bd0878` on PHP 8.4:

- Full local suite: 1,027 passed, 4 risky, 3 skipped; 3,381 assertions; no failures.
- Added regression tests for failed bindings during stack restoration and both event-assertion bugs. All three reproduced failures before their fixes.
- Local PHPStan and full Pint checks passed. PHPStan used a 1 GB memory limit.
- GitHub CI at `3bd0878`: Tests - PHP 8.4, Static Analysis, and Code Style all passed ([run](https://github.com/NativePHP/mobile-air/actions/runs/34714358681)).
- No new device/simulator verification in this pass.

## Follow-ups

- #383: optimize `native:model` lookup and improve missing-property feedback.
- #384: cache trait/reflection work.
- #344: dependent native event delivery work.


simonhamp (approved):
A bit too big, separate these kinds of things into distinct PRs in future. I agree that the breaking changes shouldn't affect anyone

PR #346, opened 2026-08-16
Stop registering view paths the bundle never ships

### LINKED ISSUE #322: view:cache crashes on every debug cold boot — package excludes a folder it still registers as a view path

Just running the app normally (`php artisan native:run android`, debug build). No custom setup — this happens on a stock install.

### What happened?

On every cold start of a debug build, the app's Laravel log shows this error (and if it happens while the user has the app open, it visibly crashes and needs a reload):

```
[ERROR] The "vendor/nativephp/mobile/src/../resources/views" directory does not exist.

Symfony\Component\Finder\Exception\DirectoryNotFoundException
at vendor/symfony/finder/Finder.php:663
#0 vendor/laravel/framework/.../ViewCacheCommand.php(88): Finder->in(Array)
#1 vendor/laravel/framework/.../ViewCacheCommand.php(43): ViewCacheCommand->bladeFilesIn(Array)
...
```

The same thing happens for a second path too:

```
[ERROR] The "vendor/nativephp/mobile/src/../resources/jump/views" directory does not exist.
```

### Why this happens

`NativeServiceProvider` registers Blade views from `vendor/nativephp/mobile/resources/views` (via Spatie's `hasViews()`) and `vendor/nativephp/mobile/resources/jump/views` (via an explicit `loadViewsFrom()` call).

But `Native\Mobile\Support\BundleExclusions::VENDOR_PATHS` lists `vendor/nativephp/mobile/resources` as excluded from every app bundle — so that whole folder, including both of those view directories, is stripped out before the app ever reaches a device.

The result: the package registers view paths that its own build process guarantees won't exist on the device. The first artisan command that runs `view:cache` (which NativePHP itself runs automatically after every fresh install/update) crashes, because Symfony's `Finder` throws if a registered path is missing.

### How to reproduce

1. Fresh Laravel 13 app, `composer require nativephp/mobile` (tested on 4.1.0)
2. `php artisan native:install android -n`
3. `php artisan native:run android` on a debug build
4. Check the app's Laravel log after the first launch — the error above appears every time (each debug cold boot re-extracts the bundle and re-runs `view:cache`)

### Suggested fix

Either:
- Don't exclude `vendor/nativephp/mobile/resources` wholesale — carve out `resources/views` and `resources/jump/views` from the exclusion, or
- Ship those two directories (even empty, with a `.gitkeep`) inside whatever `BundleExclusions` treats as "required," the same way `REQUIRED_DIRECTORIES` already does for `storage/framework/views` etc.

### Related

#245 looks like the same root-cause pattern (bundler excludes a folder that something else still needs at runtime), just a different folder.

---

Loving what you guys are doing and the direction NativePHP is going in 🙌

gwleuverink:
Fixes #322.

Every debug boot runs `view:cache` on the device, and it walks every registered view path with a Symfony Finder that throws when a path is missing. Two of the provider's registrations point at directories no bundle ever contains, because all three packers strip `vendor/nativephp/mobile/resources` (11MB of project templates the runtime never reads). So every cold boot logs this and dies:

```
The ".../vendor/nativephp/mobile/src/../resources/views" directory does not exist.
```

The two registrations are `->hasViews()` and the `jump` namespace. Neither serves a purpose at runtime. `hasViews()` registers the package root `resources/views`, which contains a single `.gitkeep`, and its publish tag (`nativephp-mobile-views`) is never published by `native:install`. As far as I can tell it's scaffolding from the Spatie package skeleton: the `.gitkeep`-only views dir and the provider call both match the template, unchanged since the Init commit. The empty dir is deleted here as well. The real component views went to `src/resources/views` instead, under a separate registration that ships fine and stays. The `jump::` namespace has zero `view()` references anywhere, the jump server reads those blades as raw text with `file_get_contents` on the host.

I checked whether anything contributes files into those two directories in a userland project before deleting the registrations. Nothing does: my test project symlinks the package into vendor, so any runtime write would show up in git status, and after many install, plugin-register and build cycles the tree is clean. mobile-ui registers no view namespace and every plugin write path resolves into the native build tree.

I actually went the other way first, shipping the two dirs as empty entries from every packer. It worked, but it put the burden on three different packers forever, and once it was clear the registrations are unused skeleton there was nothing left to ship around. Two deleted lines in the provider instead.

While hardening the regression test I found a third leg of the same crash: the provider also registers `resources/views/native` as a view location for every app, and only `native:make-component` ever creates that dir. An app that never scaffolded a native component hits the same `view:cache` crash, after `view:clear` has already run, so it's left with no compiled views at all. That location is genuinely wanted (it's where an app's own native views live), so this leg is fixed the way the issue originally suggested: `resources/views/native` joins `BundleExclusions::REQUIRED_DIRECTORIES` and ships as an empty entry, which every packer already honors.

The provider has a comment documenting a deliberate choice to register view paths unconditionally so IDE plugins that scan providers statically can index them. That argument holds for `resources/views/native` and the `src/resources/views` namespace, both of which stay exactly as they are, but an IDE gains nothing from indexing a `.gitkeep` or a namespace no code references. If I'm missing a reason either registration needs to stay, @simonhamp @shanerbaner82, happy to hear it.

## Verification

- A new test locks the invariant down: every view path the package registers must exist and must survive the bundle excludes. Paths are normalized lexically (Flysystem's path normalizer), because Spatie registers the raw `src/../resources/views` string when the dir is missing and `realpath()`-based checks go blind on it. Proven red for both regression shapes: re-adding `hasViews()` and re-adding the jump registration.
- A second test pins `nativephp-mobile::components.native-element-with-children` staying resolvable, since it rides the registration that stays. The release bundle test additionally asserts `resources/views/native/` ships.
- Full suite green, pint clean, and `view:cache` on the host exits clean.
- Android emulator, fresh build and forced cold boot: the bundle contains no `vendor/nativephp/mobile/resources` entries, logcat shows `view:cache` running post-extraction and producing 66 compiled views, and the `DirectoryNotFoundException` count in the persisted log stayed frozen at the 5 pre-fix entries. The app renders normally.
- The third leg, proven in the failing shape: built the test app with `resources/views/native` deleted from its source. The bundle ships it as an empty entry, the dir lands on device, `view:cache` runs clean, and the exception count stays frozen. That exact build crashed on every boot before this change.
- iOS simulator: `app.zip` contains no vendor resources entries, ships `resources/views/native/`, and the app boots and renders.
- Ran `native:jump` against the branch: the QR page renders and serves normally, since the jump server reads its blades directly off disk and never used the removed namespace.

### DIFF
--- a/src/NativeServiceProvider.php
+++ b/src/NativeServiceProvider.php
@@ -60,7 +60,6 @@ class NativeServiceProvider extends PackageServiceProvider
         $package
             ->name('nativephp-mobile')
             ->hasConfigFile('nativephp')
-            ->hasViews()
             ->hasRoute('api')
             ->hasCommands([
                 PackageCommand::class,
@@ -195,8 +194,10 @@ class NativeServiceProvider extends PackageServiceProvider
     {
         parent::boot();
 
+        // Only src/resources/views is registered. The package root
+        // resources/ dir never ships in the app bundle, and the
+        // device's view:cache throws on any missing path.
         $this->loadViewsFrom(__DIR__.'/resources/views', 'nativephp-mobile');
-        $this->loadViewsFrom(__DIR__.'/../resources/jump/views', 'jump');
 
         // Register `resources/views/native` as a primary view-finder
         // location (mirrors Livewire's `resources/views/livewire`
--- a/src/Support/BundleExclusions.php
+++ b/src/Support/BundleExclusions.php
@@ -63,6 +63,11 @@ class BundleExclusions
         'storage/framework/cache',
         'storage/framework/sessions',
         'storage/framework/views',
+        // The provider registers this as a view-finder location for every
+        // app, and the device's view:cache throws when it is missing.
+        // Only native:make-component creates it, so apps without a
+        // scaffolded component must still ship the empty dir.
+        'resources/views/native',
     ];
 
     /** Non-runtime patterns matched only inside vendor packages. */
--- a/tests/Feature/ReleaseBuildBundleTest.php
+++ b/tests/Feature/ReleaseBuildBundleTest.php
@@ -92,6 +92,11 @@ class ReleaseBuildBundleTest extends TestCase
         $this->assertFalse($zip->statName('bootstrap/cache/packages.php'));
         $this->assertNotFalse($zip->statName('bootstrap/cache/'));
 
+        // The provider registers this location unconditionally, so the
+        // bundle must ship it even when the app never scaffolded a
+        // native component, or view:cache throws on device (#322).
+        $this->assertNotFalse($zip->statName('resources/views/native/'));
+
         // Delegation to BundleFileManager, asserted by outcome: config
         // excludes and vendor slimming patterns shape the final zip.
         $this->assertNotFalse($zip->statName('app/Example.php'));
--- a/tests/Feature/ViewRegistrationTest.php
+++ b/tests/Feature/ViewRegistrationTest.php
@@ -1,0 +1,26 @@
+<?php
+
+use Illuminate\Support\Str;
+
+// The bundle packers strip the package root resources/ dir, and the
+// device's view:cache walks every registered path with a Finder
+// that throws when one is missing. So no registration may
+// ever point inside that directory (#322).
+it('registers no view paths under the unshipped package resources directory', function () {
+    $packageResources = realpath(__DIR__.'/../../resources');
+
+    $finder = app('view')->getFinder();
+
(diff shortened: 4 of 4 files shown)


shanerbaner82 (approved):
Looks good. view:cache crash on missing package view paths is real, and dropping the unused hasViews()/jump:: registrations is the right fix instead of shipping empty dirs forever. Shipping empty resources/views/native covers apps that never ran native:make-component. Tests lock both shapes. Jump still reads those blades off disk, so the namespace drop is safe.

PR #351, opened 2026-08-17
Make optional route segments resolvable

sadiqk2:
### The bug

```php
Route::native('/posts/{slug?}', PostScreen::class);
```

This route cannot match any URI at all — not `/posts/hello`, not `/posts`.

`NativeRouter::resolve()` compiles its pattern with `preg_replace('/\{(\w+)\}/', '(?P<$1>[^/]+)', $pattern)`, and `\w` does not match `?`. The `{slug?}` placeholder therefore survives into the regex verbatim, so the compiled pattern only ever matches the literal string `/posts/{slug?}`.

What turns a dead pattern into a device-visible failure: `BootPlanner` on both Android and iOS matches these patterns when it decides whether to boot native. An app using the documented optional syntax boots into the native runloop and *then* resolves to no screen — a blank native window instead of the WebView fallback that would otherwise have carried it.

### The fix

- Optional segments are substituted first, and the pattern consumes the leading slash with them (`#/\{(\w+)\?\}#` → `(?:/(?P<$1>[^/]+))?`), so the segment and its separator disappear together and `/posts` matches.
- An omitted segment is dropped from `params` rather than reported as `''`. A screen checking whether it received a parameter should not be told yes for a segment nobody sent. `[^/]+` cannot match empty, so no legitimate value is filtered.

Required segments are unaffected.

### Verification

New `tests/Unit/Edge/NativeRouterOptionalParamsTest.php` — 6 cases: segment present, segment omitted, `isNativeRoute()` agreeing with `resolve()` for both (the BootPlanner half), required segments still required, an optional segment after a required one, and an optional segment not swallowing a deeper path.

4 of the 6 fail against `main`. Full suite: 887 passed, and the 13 failures are identical to `main`'s baseline (Android splash-screen and release-build tests, unrelated to this change). `vendor/bin/pint --test` clean.

### DIFF
--- a/src/Edge/NativeRouter.php
+++ b/src/Edge/NativeRouter.php
@@ -191,11 +191,26 @@ class NativeRouter
 
         // Pattern match with route parameters
         foreach (static::$routes as $pattern => $entry) {
-            $regex = preg_replace('/\{(\w+)\}/', '(?P<$1>[^/]+)', $pattern);
+            // Optional segments FIRST, and consuming their leading slash: `\w` does not
+            // match `?`, so a `{slug?}` placeholder previously survived into the regex
+            // literally and the pattern could only ever match the string "{slug?}".
+            // BootPlanner on both platforms matches these patterns happily, so a route
+            // using the documented `{param?}` syntax booted into the native runloop and
+            // then resolved to no screen at all — /posts/{slug?} matched neither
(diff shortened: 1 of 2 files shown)


shanerbaner82:
#343 closes this one

PR #356, opened 2026-08-18
Make text whitespace symmetric with the browser, with whitespace-* classes to opt out

### LINKED ISSUE #336: [Masterclass] <text> slot content is whitespace-normalized, destroying line breaks in user-generated text

Rendering a user-written message that contains blank lines between paragraphs - `<text>{{ $message }}</text>` - and having the paragraph breaks show up.

### What happened?

Expected: the message renders with its line breaks intact.

Actual: it renders as one run-on line. All the newlines are gone.

`Text.php:20` runs slot content through `preg_replace('/\s+/', ' ', ...)`.
Because `\s` matches newlines, `\n\n` collapses to a single space.

The `:text` attribute path skips that regex and renders correctly - so the same element with the same value gives opposite results depending on whether the value arrives as an attribute or as slot content.

The data itself is fine: verified stored in the database and returned by the API with newlines intact. It's lost at render.

### How to reproduce the bug

Save a message containing a blank line, then render it both ways:

```
{{-- Slot: newlines collapse --}}
<text>{{ $message }}</text>

{{-- Attribute: newlines preserved --}}
<text :text="$message" />
```

With `$message` set to something like `"First paragraph.\n\nSecond paragraph."`, the first renders as a single run-on line and the second keeps the break.

### Debug Output

```
   INFO  NativePHP Mobile.  

+--------------------+----------+
| Package Version    | dev-main |
| PHP Version (Host) | 8.4.18   |
| OS                 | Darwin   |
| OS Version         | 25.5.0   |
| Embedded PHP       | 8.4.24   |
+--------------------+----------+

   INFO  Installed Plugins.  

+---------------------------------+----------+
| Package                         | Version  |
+---------------------------------+----------+
| nativephp/mobile-secure-storage | 1.0.2    |
| nativephp/mobile-ui             | dev-main |
+---------------------------------+----------+

   INFO  Development Tools.  

```

macOS

iOS (Device)

### Notes

Not blocking, since `:text` works today - but the difference isn't discoverable from the docs.

gwleuverink:
Fixes #336.

The report: `<text>{{ $msg }}</text>` collapses newlines while `:text="$msg"` preserves them, so user-generated text renders differently depending on which authoring form you picked. Reproduced identically on both platforms, and across every form: plain `<text>`, `<native:text>`, `<x-native-text>`, slot and attribute.

Plain Blade never touches whitespace, I checked: slot and prop come through byte-identical. The browser collapses at paint time via CSS white-space, so nothing is ever destroyed and the two sources can't diverge. EDGE collapsed at capture time in PHP instead, and only on the slot path, which is both destructive and asymmetric.

This PR makes the whole thing behave like the web. There is one whitespace policy, applied where slot and attribute text merge, and it treats both sources identically. The default collapses, exactly what CSS white-space normal does. The new whitespace-pre-line and whitespace-pre-wrap classes opt out, and whitespace-normal exists to say the default explicitly. It's pre-wrap rather than pre on purpose: native text always soft-wraps, and per the CSS table pre means no wrapping while pre-wrap means preserve and wrap, which is exactly what this renders. Verified against MDN and measured in a real browser before naming it. In practice:

```blade
{{-- $msg = "Line one\n    Line two" --}}

<text>{{ $msg }}</text>
<text :text="$msg" />
{{-- both render: Line one Line two --}}

<text class="whitespace-pre-line">{{ $msg }}</text>
<text class="whitespace-pre-line" :text="$msg" />
{{-- both render:
     Line one
     Line two          (newlines kept, indentation collapsed) --}}
```

That default is also what keeps hand-wrapped template prose working, the same way a `<p>` does:

```blade
<text>
    A long sentence that was wrapped
    over two lines in the editor.
</text>
{{-- renders as one flowing line --}}
```

Nested runs inherit the closest classed ancestor the way white-space cascades, a run's own class wins, and runs collapse without trimming, the browser trims at block boundaries only. So inline separators keep their spacing:

```blade
<text>Docs<text :text="' / '" />Guides</text>
{{-- renders: Docs / Guides --}}
```

Making every form identical also surfaced two pre-existing bugs. A self-closing `<text />` nested inside a paired `<text>` used to escape the run system entirely and render as a misplaced sibling; it's a proper in-order run now. And `<x-native-text>` had the same escape when nested, which also meant it couldn't inherit a parent's whitespace class; it now runs the exact frame cycle a precompiled `<text>` does. Parity across all seven authoring forms, nested and top-level, prop and slot, is pinned by matrix tests that name the diverging form when they fail.

The policy runs at capture time in PHP, not paint time on the device. Paint time is the browser's model, but SwiftUI and Compose render strings verbatim, so paint time here would mean implementing CSS collapsing twice in Swift and Kotlin plus a wire prop plus cross-run edge logic in both renderers. The split lands where it belongs: content shaping in PHP, layout behaviour native. That's also why whitespace-nowrap, whitespace-pre and break-spaces are left out, all three need a native no-wrap prop before they can exist truthfully.

The breaking part: `:text` no longer preserves newlines by default, and that's what people rely on today to keep them. The migration is one class:

```blade
{{-- before: :text preserved newlines as a side effect --}}
<text :text="$message" />

{{-- after: say it explicitly --}}
<text :text="$message" class="whitespace-pre-wrap" />
{{-- or, for chat-style user content: --}}
<text :text="$message" class="whitespace-pre-line" />
```

The classes and the flip ship together, so the escape hatch exists the moment the default changes. Worth a changelog callout.

Known siblings, deliberately untouched: button labels keep their own inline collapse in the precompiler, programmatic `Text::make()->text(...)` ships strings verbatim with the classes documented as a Blade-only capture policy, and EDGE elements inside a plain Blade component slot emit ahead of the component's wrapper, a pre-existing capture-order quirk. The whitespace policy resolves correctly through such slots (tested); the placement itself, including the inverse case of a native `<text>` inside an `<x-native-text>` slot, predates this branch byte-for-byte and is pinned consciously where tests touch it.

Verification. 21 tests cover every class on every source and form, run inheritance and per-run override, separator runs, nbsp (which must never collapse, and doesn't, in either CSS or this implementation), delivery through Blade component slots, the exact byte shape of the defaults, and the seven-form parity matrix; the original set was proven red on main. Full suite green. On device: a 16-case repro screen rendered identically on the Android emulator and iOS simulator, checked row by row against a plain HTML page rendering the same strings under real CSS white-space.

### DIFF
--- a/src/Edge/Components/Native/Text.php
+++ b/src/Edge/Components/Native/Text.php
@@ -16,11 +16,13 @@ class Text extends NativeBladeComponent
     public function render(): \Closure
     {
         return function (array $data) {
-            $attrs = $data['attributes']->getAttributes();
-            $text = preg_replace('/\s+/', ' ', trim(html_entity_decode(strip_tags($data['slot']->toHtml()), ENT_QUOTES, 'UTF-8')));
-            if ($text !== '') {
-                $attrs['text'] = $text;
-            }
+            // Merge slot and attribute text through the collector's shared
+            // whitespace policy so this component and the precompiled
+            // <text> tag resolve their text prop identically.
+            $attrs = NativeElementCollector::mergeSlotText(
+                $data['attributes']->getAttributes(),
+                $data['slot']->toHtml()
+            );
 
             if (NativeElementCollector::isStreaming()) {
                 NativeElementCollector::leafStreaming($this->elementType(), $attrs);
--- a/src/Edge/Components/Native/Text.php
+++ b/src/Edge/Components/Native/Text.php
@@ -16,19 +16,15 @@ class Text extends NativeBladeComponent
     public function render(): \Closure
     {
         return function (array $data) {
-            // Merge slot and attribute text through the collector's shared
-            // whitespace policy so this component and the precompiled
-            // <text> tag resolve their text prop identically.
-            $attrs = NativeElementCollector::mergeSlotText(
-                $data['attributes']->getAttributes(),
-                $data['slot']->toHtml()
-            );
+            // Run the same frame cycle as a precompiled <text> tag, so a
+            // nested component becomes a run of its parent, inheriting
+            // its whitespace mode, while a top-level one merges slot
+            // and attribute text through the identical policy. The
+            // close also emits streaming-aware, like any <text>.
+            NativeElementCollector::textOpen($data['attributes']->getAttributes());
+            echo $data['slot']->toHtml();
 
-            if (NativeElementCollector::isStreaming()) {
-                NativeElementCollector::leafStreaming($this->elementType(), $attrs);
-            } else {
-                NativeElementCollector::leaf($this->elementType(), $attrs);
-            }
+            NativeElementCollector::textClose();
 
             return '';
         };
--- a/src/Edge/Components/Native/Text.php
+++ b/src/Edge/Components/Native/Text.php
@@ -6,6 +6,11 @@ use Native\Mobile\Edge\NativeElementCollector;
 
 class Text extends NativeBladeComponent
 {
+    /**
+     * Skips the base withAttributes emission, which also means this
+     * component's frame opens AFTER its slot has rendered, unlike
+     * containers whose element opens before the slot runs.
+     */
     protected bool $handlesCollectorManually = true;
 
     protected function elementType(): string
@@ -21,6 +26,11 @@ class Text extends NativeBladeComponent
             // its whitespace mode, while a top-level one merges slot
             // and attribute text through the identical policy. The
             // close also emits streaming-aware, like any <text>.
+            //
+            // Blade closed the slot-capture buffer before this closure
+            // runs, so the echoed HTML is inert text and textOpen's
+            // buffer handoff sees the enclosing frame, never the
+            // slot's. The cycle depends on that ordering.
             NativeElementCollector::textOpen($data['attributes']->getAttributes());
             echo $data['slot']->toHtml();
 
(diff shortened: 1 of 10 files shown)


gwleuverink:
If you want to see the difference on a device rather than in the diff: I pushed the repro app at https://github.com/trailhead-labs/air-repro. Clone it, `composer update`, build, and you're looking at the before state. Swap the `nativephp/mobile` requirement to `dev-fix/336-whitespace-classes as 4.2.0`, `composer update`, rebuild, and the same 16-row screen shows the after. The README has the full steps, and `/whitespace-web` serves the same strings under real CSS white-space for a browser side-by-side.

gwleuverink:
The rest of the white-space family (nowrap, pre, break-spaces) is left out on purpose. The three shipped solve the reported problem, happen to be the part PHP can own. 

wrapping included: pre-wrap is named pre-wrap because native text always wraps, and CSS pre means no wrapping. 

The other three all need a native no-wrap prop first, and I can't really make a case for them: nowrap's job on native is truncation (truncate / line-clamp-* would be the better follow-up), and break-spaces can't be done faithfully on either platform.

shrutibalasawebdev:
Verified on Android and iOS with a real chat app. Line breaks in my chat bubbles stopped working, and adding `whitespace-pre-line` brought them back.

shanerbaner82:
Correction to what I wrote a moment ago: the force-push to this branch was refused (the fork rejects maintainer pushes from my account over HTTPS and SSH), so this branch is unchanged.

The rebased version lives in https://github.com/NativePHP/mobile-air/pull/425 instead: your 11 commits unchanged with original authorship, plus one leading commit that reverts #424, which landed in the meantime with a narrower version of the same fix. Full suite passes there (959 tests), Pint clean, no conflicts. That PR supersedes this one; #424's conservative `:text` default was the wrong call and your approach is what we want. Sorry for the churn.

shanerbaner82:
Closing: this landed on main via #425, which is this branch rebased with #424 reverted in front. Your commits are in with original authorship. Thanks.

PR #357, opened 2026-08-18
Make the iOS hot-reload port configurable

shanerbaner82:
## Why

The in-app hot-reload server bound a hardcoded `9999`. That port is **host-wide** — a simulator shares the host's localhost, and physical devices are tunnelled to the same host port by iproxy — so only one app could hot-reload at a time.

Worse, the collision is silent. When a second app (or a stale instance of the same one) already holds the port, `triggerIosReload()`'s `fsockopen` connects successfully and reports nothing wrong. The trigger lands on the wrong listener, so files sync, watchman reports success, and the live app never reloads.

## What

Adds `nativephp.hot_reload.port`, defaulting to `9999` — no behaviour change for existing apps.

- `HasHotReloadPort` — one `hotReloadPort()` accessor, used by `WatchesIos`, `RunsIos`, and `BuildIosAppCommand`
- `WatchesIos` — reload trigger, iproxy forward (`{port}:{port}`), both `lsof` cleanups, and the failure message
- `RunsIos` — the pre-launch port free
- `BuildIosAppCommand` — writes `NATIVEPHP_HOT_RELOAD_PORT` into Info.plist, alongside the existing `BIFROST_APP_ID` handling
- `HotReloadServer.swift` — reads that key from `Bundle.main`, falling back to 9999

Swift can't read the Laravel config, so the port has to be baked into Info.plist at build time and read back at launch. Changing the port therefore needs a rebuild — noted in the config comment.

**iOS only.** Android signals a reload by pushing a file into the app's storage over adb and never binds a port, so it ignores the key.

## Tests

`tests/Feature/IosHotReloadPortTest.php`, modelled on `IosInterfaceStyleTest`: the configured port is written, an unset port falls back to the default, and an existing key is updated rather than duplicated.

The fallback test caught a real bug during development — `config()` only applies its default when the key is *absent*, so an app that published the key but left it empty yielded port `0`. `hotReloadPort()` now coerces empty to the default.

### DIFF
--- a/config/nativephp.php
+++ b/config/nativephp.php
@@ -386,6 +386,20 @@ return [
     |--------------------------------------------------------------------------
     */
     'hot_reload' => [
+        /*
+         * TCP port the in-app hot-reload server listens on.
+         *
+         * The port is host-wide — a simulator shares the host's localhost, and
+         * physical devices are tunnelled to the same host port by iproxy — so
+         * give each app its own port if you want two of them hot-reloading at
+         * the same time. Baked into Info.plist at build time; changing it
+         * needs a rebuild.
+         *
+         * iOS only. Android signals a reload by pushing a file into the app's
+         * storage over adb, so it never binds a port and ignores this value.
+         */
+        'port' => 9999,
+
         'watch_paths' => [
             'app',
             'resources',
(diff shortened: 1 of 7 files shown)


simonhamp:
Closing in favor of #482

PR #369, opened 2026-08-24
feature: let the app declare which languages it supports

SRWieZ:
## The gap

Which languages an app supports is the app developer's call. Today it is inferred from the keys of `permission_localizations` — the app's own **and every installed plugin's**. A plugin translating its camera explainer into ten languages therefore speaks for the app:

- on **Android**, nothing happened at all: 13+ only lists an app in the per-app language picker when its manifest points at a locale-config resource, and the compiler never wrote one. A fully localized NativePHP app followed the system language but could never be switched per app.
- on **iOS**, the opposite problem. The locales were registered (per-locale `InfoPlist.strings`, `knownRegions`) — including the plugins' — and the App Store product page derives its **Languages** list from the `.lproj` folders in the binary, not from App Store Connect metadata. So an English-only app installing a well-translated plugin advertises ten languages on the store.

| Before | After |
|:---:|:---:|
|  |  |
| No **Language** entry in App info | **Language** entry appears |

## The fix

Reshaped from the first version of this PR, which derived the list from `permission_localizations`, after @gwleuverink and @simonhamp pointed out that plugins shouldn't drive it.

A new `supported_locales` declares the languages, and it is the only input to them:

```php
// config/nativephp.php
'supported_locales' => ['fr', 'nl'],
```

Both compilers resolve that one list through `SupportedLocales`:

- `config('app.locale')` always leads it, so users can explicitly switch back to the base language (this replaces the `default_locale` key the first version added — @simonhamp, this is your inline comment);
- entries are normalized (`nl_NL` → `nl-NL`), de-duplicated and validated before they reach an XML attribute or a plist;
- **plugin localizations are content, never scope** — a plugin's `ios.info_plist_localizations` entry is written only for a language the app already supports, and a plugin declaring `CFBundleLocalizations` in its `info_plist` is ignored;
- declaring nothing means supporting one language, so an app that declares nothing is left completely untouched.

**Android** writes `res/xml/locales_config.xml` and injects `android:localeConfig` into `<application>`. Worth noting AGP's `generateLocaleConfig` can't do this job: it scans `res/values-*`, which is empty in a NativePHP app, and it can't coexist with a hand-written locale-config.

**iOS** declares `CFBundleLocalizations`. That's Apple's mechanism for an app whose translations don't live in `.lproj` — which is every NativePHP app, since they translate in PHP against Laravel's `lang/` files. It is what makes the app offer a language in Settings and what the store lists. The `.lproj` folders stay for what they're actually for: the translated permission strings.

## Removal works now too

@gwleuverink's five-step scenario, fixed in both directions. Both paths run **before** their compiler's no-plugins early return — the exact shape of "the developer removed everything" — and undo their own work:

- Android deletes the xml and strips the manifest attribute;
- iOS rewrites `CFBundleLocalizations` (set, not merged — a merge unions lists by content, so a language could never be un-shipped), deletes stale `.lproj` folders and prunes their `knownRegions` entries.

A `.lproj` holding anything beyond the generated `InfoPlist.strings` is never deleted: that folder belongs to the developer.

## Compilers warn instead of guessing

Scanning `lang/` to decide what ships would offer a half-translated language the moment someone starts one, and Laravel publishes `lang/en/validation.php` into apps that aren't localized at all. So the declarations stay explicit, and the compiler names the mismatch instead:

> Translations exist for de but nativephp.supported_locales does not list them, so the app will not offer them

## ⚠️ Breaking

An app relying on a plugin's translations now has to declare that locale to keep them. One line, and the resulting behaviour is the one the developer actually wanted — worth a note in the release notes.

## Verified on an emulator (Android 16)

The picker lists exactly the declared locales, and selecting one changes what the app's PHP sees (`Locale.getDefault()` → `Device::getInfo()['language']`):

| Picker | App in French |
|:---:|:---:|
|  |  |

## Tests

`SupportedLocalesTest` pins the resolution rules without standing up a fake Gradle or Xcode project. The compiler tests cover both platforms: declared locales are written, plugin-only locales aren't, removal clears everything, recompiling is idempotent, invalid entries warn, a developer-owned `.lproj` survives, and `knownRegions` entries are removed whole (dropping `nl` leaves `nl-NL` alone). Full suite green.

### DIFF
--- a/config/nativephp.php
+++ b/config/nativephp.php
@@ -156,6 +156,20 @@ return [
     |
     */
 
+    /*
+    |--------------------------------------------------------------------------
+    | Default Locale
+    |--------------------------------------------------------------------------
+    |
+    | The language your unlocalized strings (Info.plist values, UI copy) are
+    | written in. It leads the Android locales_config.xml generated when you
+    | declare `permission_localizations`, so users can always switch back to
+    | it explicitly in the per-app language picker.
+    |
+    */
+
+    'default_locale' => 'en',
+
     'permission_localizations' => [
         // 'nl' => [
         //     'NSCameraUsageDescription' => 'Gebruikt om een profielfoto te maken.',
--- a/src/Plugins/Compilers/AndroidPluginCompiler.php
+++ b/src/Plugins/Compilers/AndroidPluginCompiler.php
@@ -187,6 +187,11 @@ class AndroidPluginCompiler
         // even when the list is empty so a removed plugin's declaration is cleared).
         $this->injectGradlePlugins($allPlugins);
 
+        // Declare the app's locales so Android 13+ offers the per-app
+        // language picker (runs even with no plugins — the app's own
+        // permission_localizations are reason enough).
+        $this->writeLocalesConfig($allPlugins);
+
         if ($allPlugins->isEmpty()) {
             $this->generateEmptyRegistration();
             $this->generateEmptyRendererRegistration();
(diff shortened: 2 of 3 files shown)


gwleuverink:
Awesome, thanks! Great PR 🚀 

One design question first, more for @simonhamp (and my own understanding) than for this PR. The language list comes from `permission_localizations` keys, including plugin ones. A plugin translating its permission explainer into ten languages would make an English-only app offer ten languages in the picker. iOS already does the same through the lproj folders, so this is consistent, but maybe the supported languages want their own key that both platforms read? Not sure on this one but it sounds like a bug 🤔 

Then a smaller one, about what happens when the localizations go away again by an update after the app is already installed:

1. Declare `fr` and `nl` under `permission_localizations`, run `native:run android`. `locales_config.xml` and the `android:localeConfig` attribute are written, the picker shows both languages.
2. Set `permission_localizations => []` (or remove the only plugin that shipped them) and run `native:run android` again.
3. `writeLocalesConfig` returns early on the empty list, so the xml and the attribute stay exactly as they were.
4. The new APK still carries `localeConfig` with `fr` and `nl`, and the picker keeps offering them to an app that no longer declares any.
5. Only a fresh `native:install` clears the manifest, and since `nativephp/` is gitignored a dev building releases from a long-lived scaffold can ship it that way.

Admittedly, this will likely not happen often in nativephp apps, but it might. It is cosmetic, but `injectPluginProguardRules` and `injectGradlePlugins` right above it do clear their block on an empty list. Was skipping that intentional, or would you be up for having this one strip the attribute and delete the file when the list is empty?

SRWieZ:
I think this was originally based on iOS, if I recall correctly. 

As a user, I’d personally prefer full control over which locales are suggested. 

It really depends more on my Laravel app’s translation strings than on plugin permissions. 

So maybe a new config ?

gwleuverink:
Exactly. We cant have a random plugin contributing a bunch of extra language options the app doesnt even support. 

Dedicated config makes sense to me. Good to be explicit. What do you think bossmen @simonhamp @shanerbaner82 ?

config/nativephp.php
```diff
@@ -156,6 +156,20 @@
...
+    |--------------------------------------------------------------------------
+    |
+    | The language your unlocalized strings (Info.plist values, UI copy) are
+    | written in. It leads the Android locales_config.xml generated when you
+    | declare `permission_localizations`, so users can always switch back to
+    | it explicitly in the per-app language picker.
+    |
+    */
+
+    'default_locale' => 'en',
```
simonhamp, line 197: Why not use Laravel's own `app.locale`?

simonhamp:
I agree on all points:

1. If we're cleaning up plugin dependencies, then we should clean up their language strings too
2. I don't think the plugin localisations should drive language options - I agree that this should be the app developer's choice

So on balance, I think this _should_ be a separate config somewhere, declaring which locales _you_ want to support in _your_ app.

Then if the plugin localisations line up with those, great!

Therefore I think this PR doesn't make sense as is. I'll leave open in case you want to reshape it into this. Up to you @SRWieZ

SRWieZ:
I'll update the PR

src/Plugins/Compilers/AndroidPluginCompiler.php
```diff
@@ -210,10 +211,10 @@ public function compile(): void
             $this->warn(LegacyFirebaseConfig::deprecationNotice($legacyFirebase, 'android'));
         }
 
-        // Declare the app's locales so Android 13+ offers the per-app
-        // language picker (runs even with no plugins — the app's own
-        // permission_localizations are reason enough).
-        $this->writeLocalesConfig($allPlugins);
+        // Declare the languages the app supports so Android 13+ offers the
```
gwleuverink, line 214: same for iOS right?

SRWieZ: yes

simonhamp:
Yep this is wayyyy better. Just need to update the docs

gwleuverink (approved):

gwleuverink:
Beautiful man! was waiting on this one before writing translations for Humm
Thankyou!!! 🙏🏻

mustafa-online:
Does this support RTL if the "Arabic" is selected?

Issue #360, opened 2026-08-20
Native events never reach webview apps on Android (appearance, shake)


gwleuverink:
### What were you trying to do?

Use native events from a webview app. Two of them, hit in the same session:

- branch on the system theme server-side with `System::appearance()` / `System::isDarkMode()`
- react to a shake with `#[On(ShakeDetected::class)]`

Found while reviewing #359 

### What happened?

Neither works on a webview screen. EDGE screens are fine, so the events themselves work.

`System::appearance()` freezes on whatever the appearance was the first time it was called and stays wrong for the life of the PHP process. The bridge reports the flip immediately, only the PHP-facing API goes stale, and there is no `forgetAppearance()` to recover with.

`ShakeDetected` never arrives at all, not in PHP and not through the JS `native-event` listener.

iOS gets appearance right. Shake looks affected there too (`ShakeDetector.swift:23` sends the same way), but I only tested that one on Android.

### How to reproduce the bug

Appearance, as the start URL, then `adb shell cmd uimode night yes`:

```php
Route::get('/appearance-web', function () {
    $api = \Native\Mobile\Facades\System::appearance();
    $os = nativephp_call('System.GetAppearance', '{}');

    return response('<meta http-equiv=refresh content=2><pre>'."api={$api} os={$os}</pre>");
});
```

Shake goes the same way: `#[On(ShakeDetected::class)]` on an EDGE screen fires, the same listener on a webview screen never does. Drive the accelerometer past the 2.7g threshold a few times to trigger it:

```
adb emu sensor set acceleration 0:35:0
adb emu sensor set acceleration 0:-35:0
```

### Debug Output

Appearance, on a webview screen:

```
$ adb shell cmd uimode night yes

bridge   System.GetAppearance  => {"appearance":"dark"}   (bridge sees the flip)
PHP      System::appearance()  => light                   (stays light, same process)
```

Shake, same build and same injection profile, only the screen type differs:

```
EDGE screen      8 shakes injected   handler fired, counter reached 11
webview screen   8 shakes injected   nothing in PHP, nothing in JS
```

### Which operating systems have you seen this occur on?

macOS

### Which platforms were you trying to build for?

Android (Emulator)

### Notes

Both events are sent with `NativeElementBridge.sendNativeEvent` (`MainActivity.kt:463` for appearance, `:611` for shake), which only feeds the EDGE element queue. The only thing draining that queue is `NativeComponent`'s render loop, so on a webview screen nothing is ever delivered.

[...]

PR #371, opened 2026-08-24
feature: OrientationChanged event + System::orientation() — react to device rotation from PHP

SRWieZ:
## What this does

When the app window changes between portrait and landscape, PHP is notified, so a screen can render a different layout for each orientation.

Until now, an app that allows both orientations would reflow natively when its window changed, but the PHP side never knew it happened. There was no way to say "in landscape, put the score tiles side by side".

## How you use it

React to an orientation change in any screen:

```php
use Native\Mobile\Events\System\OrientationChanged;

#[On(OrientationChanged::class)]
public function rotated(string $orientation): void
{
    // 'portrait' or 'landscape' — state changed, screen re-renders
}
```

Or ask at any time:

```php
System::orientation();  // 'portrait' | 'landscape'
System::isLandscape();  // bool
System::isPortrait();   // bool
```

## How it works

It follows the existing `AppearanceChanged` pattern end to end:

- **Native pushes an event when the app window's aspect changes.**
  - Android emits from `onConfigurationChanged`. The last orientation lives in the companion object, so an activity recreation caused by multi-window resizing compares the fresh configuration in `onCreate()` and emits when PHP's process cache would otherwise remain stale.
  - iOS watches layout changes but reads the actual app-window bounds for both push and query paths. Keyboard safe-area changes therefore cannot create false iPad orientation events. It uses the current iOS 17+ `onChange` form.
- **A listener in `NativeServiceProvider` keeps `System`'s cached value fresh**, so reads avoid another bridge round-trip after the first one.
- **`System.GetOrientation` backs the first read**, before any orientation event has arrived.

## Good to know

- Orientation describes the **app window**, not necessarily the physical device. They can differ in iPad and Android multi-window modes.
- The event only fires if the app enables a second orientation in `config/nativephp.php`.
- The payload is deliberately minimal: `'portrait' | 'landscape'`.
- Because the event implements `BroadcastsGlobally`, it also reaches app-wide `Event::listen()` listeners.
- Android webview screens depend on [#379](https://github.com/NativePHP/mobile-air/pull/379) to route native events into PHP.
- The documentation site needs a follow-up orientation page; it does not live in this repository.

## Tests

- Orientation event/global-dispatch and cache behaviour.
- Invalid-value guard and native-payload rebuild.
- Explicit service-provider listener coverage for both orientation and appearance events.

Full suite: 902 passed. Pint and PHPStan clean.

### DIFF
--- a/resources/androidstudio/app/src/main/java/com/nativephp/mobile/bridge/BridgeFunctionRegistration.kt
+++ b/resources/androidstudio/app/src/main/java/com/nativephp/mobile/bridge/BridgeFunctionRegistration.kt
@@ -29,6 +29,7 @@ fun registerBridgeFunctions(activity: FragmentActivity, context: Context) {
     // plugin). iOS twin: Bridge/Functions/SystemFunctions.swift.
     registry.register("System.OpenAppSettings", SystemFunctions.OpenAppSettings(context))
     registry.register("System.GetAppearance", SystemFunctions.GetAppearance(context))
+    registry.register("System.GetOrientation", SystemFunctions.GetOrientation(context))
     registry.register("System.MinimizeApp", SystemFunctions.MinimizeApp(activity))
 
     // Dialog.* — core built-in (migrated from the nativephp/mobile-dialog
--- a/resources/androidstudio/app/src/main/java/com/nativephp/mobile/bridge/functions/SystemFunctions.kt
+++ b/resources/androidstudio/app/src/main/java/com/nativephp/mobile/bridge/functions/SystemFunctions.kt
@@ -80,7 +80,8 @@ object SystemFunctions {
     }
 
     /**
-     * Current screen orientation (portrait / landscape). Backs
+     * Current app-window orientation (portrait / landscape), which may differ
+     * from the physical device orientation in multi-window mode. Backs
      * `System::orientation()` / `isLandscape()` for the cold read before the
      * first OrientationChanged push.
      * Returns:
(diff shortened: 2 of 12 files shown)


gwleuverink (changes requested):
Great addition, this one was missing. Verified on an Android emulator and the iOS simulator: works as described on EDGE screens, one event per rotation, clean cold read, no false positives on theme flips.

Two cases where it goes wrong, both measured:

**iPad keyboard.** With `ipad => true`, focusing a text field in portrait emits `landscape`, hiding the keyboard emits `portrait`. The GeometryReader measures the safe-area frame, which the keyboard shrinks below square on an iPad (1024x1366 → 1024x963); `GetOrientation` reads `window.bounds`, so the two disagree while the keyboard is up, and a rotation with the keyboard open leaves `System::orientation()` stuck on landscape. Read the window (or `interfaceOrientation`) on both sides, or put `.ignoresSafeArea(.keyboard)` on the reader. iPhone is unaffected.

**Android multi-window.** A resize that changes `smallestScreenSize` or `screenLayout` recreates the activity instead of calling `onConfigurationChanged` (the manifest only declares `uiMode|colorMode|orientation|screenSize`), so nothing is pushed and `onCreate` re-seeds `lastOrientation` silently. PHP keeps the stale value for the rest of the process. Keeping `lastOrientation` in the companion object and emitting from `onCreate` when the fresh read differs covers it.

Smaller:

- Docblock: both platforms measure the window aspect, not the device. Worth saying, and the Swift `GetOrientation` comment "same signal the push uses" is not true.
- `.onChange(of:) { size in }` is the iOS 17-deprecated form (target is 18.2; the colorScheme twin has it too).
- Nothing covers the provider listener: deleting the `Event::listen` leaves the suite green for both events.
- Docs need a follow-up, orientation has no page yet.

Merge after #379: on Android a rotation on a webview screen never reaches PHP (the #360 gap, inherited from `AppearanceChanged`) and the cache then stays wrong process-wide. #379 routes those events through `/_native/api/events`, which dispatches through the provider listener, so that closes on its own.

gwleuverink (approved):
Perfect 🙌🏻

PR #374, opened 2026-08-24
fix: scrolling gets stuck on elements with press-* feedback (iOS 18)

SRWieZ:
## What's wrong
**On iOS 18, a scroll that starts on any element with `press-scale` (or any `press-*` prop) never scrolls.** The press feedback watched for touch contact with a zero-distance `DragGesture` attached via `simultaneousGesture`; on iOS 18 that drag claims the touch before the surrounding ScrollView can begin its pan. On a screen made mostly of pressable cards, the only way to scroll is to aim for the gaps between them.

Minimum repro (fill a screen with these — swipes that start on a card don't scroll):

```blade
<native:scroll-view class="w-full flex-1">
    <native:column class="w-full p-5 gap-3">
        @for ($i = 0; $i < 20; $i++)
            <native:pressable :press-scale="0.97" class="w-full h-[80] rounded-2xl bg-white items-center justify-center">
                <native:text font="semibold" class="font-semibold">Card {{ $i }}</native:text>
            </native:pressable>
        @endfor
    </native:column>
</native:scroll-view>
```

## What this does
- Swaps the drag for `_onButtonGesture(pressing:)` — the recognizer SwiftUI's own `Button` uses. A ScrollView already knows how to share touches with buttons: pans win, taps still land, and the pressed effect engages only once the system resolves the touch as a press.
- `@tap` is untouched (it fires through its own handler). Android is untouched.

No screenshot — the bug is a gesture, not pixels; the repro above shows it in seconds on an iOS 18 device.

### DIFF
--- a/resources/xcode/NativePHP/NativeRender/NodePressFeedbackModifier.swift
+++ b/resources/xcode/NativePHP/NativeRender/NodePressFeedbackModifier.swift
@@ -9,10 +9,16 @@ import SwiftUI
 ///   - `press-opacity`      — opacity while pressed (e.g. 0.7).
 ///   - `press-translate-y`  — Y offset while pressed (points).
 ///
-/// Press detection uses a zero-distance `DragGesture` attached via
-/// `simultaneousGesture` so it composes with the existing tap and
-/// long-press handlers — the visual feedback happens immediately on
-/// press-in, `@tap` still fires on tap, and scrolls aren't blocked.
+/// Press detection uses SwiftUI's button recognizer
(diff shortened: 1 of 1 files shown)


shanerbaner82 (approved):
iOS 18 scroll-vs-press is a real bug. Switching press highlight to the button recognizer so ScrollView can still pan is the right trade. Note this uses `_onButtonGesture`, which is underscored, but the public DragGesture path is what made screens unscrollable.

shanerbaner82:

PR #375, opened 2026-08-24
fix: Android bottom-bar draws under the system navigation bar

SRWieZ:
## What's wrong
On Android, `<native:bottom-bar>` content draws **underneath the system navigation bar** (gesture pill or 3-button strip). Material's own `NavigationBar` pads itself above the system bar, but the stack renderer drops custom bottom-bar content into the Scaffold slot raw — no insets — so it sits flush with the window edge, and the home button lands on top of your buttons.

Minimum repro:

```blade
<native:bottom-bar>
    <native:row class="w-full items-center justify-between px-5 py-3 bg-white">
        <native:column class="w-12 h-12 rounded-full bg-zinc-200 items-center justify-center"><native:text font="bold" class="font-bold">1</native:text></native:column>
        <native:pressable class="px-8 h-[44] rounded-full bg-blue-600 items-center justify-center"><native:text font="semibold" class="font-semibold text-white">Primary</native:text></native:pressable>
        <native:column class="w-12 h-12 rounded-full bg-zinc-200 items-center justify-center"><native:text font="bold" class="font-bold">2</native:text></native:column>
    </native:row>
</native:bottom-bar>
```

## What this does
- Pads the bottom-bar slot with the navigation-bar inset and paints the bar's own background color through the inset — the same "bleed" treatment the iOS renderer gives the home-indicator strip, so the strip matches the bar instead of showing the window background.
- The IME inset is excluded: the root `imePadding` already lifts the bar above the keyboard, so keeping the nav-bar inset there would float it too high.
- The tabs renderer needs no change (its bottom bar already sits inside the Scaffold padding). No iOS or PHP changes.

## Before / after (Android emulator, API 36, 3-button navigation, same Blade)
Before: the bar is flush with the window edge — the home button sits ON the Primary button. After: the bar rides above the nav strip and its background bleeds through it.

### DIFF
--- a/resources/androidstudio/app/src/main/java/com/nativephp/mobile/ui/nativerender/NativeRootStackRenderer.kt
+++ b/resources/androidstudio/app/src/main/java/com/nativephp/mobile/ui/nativerender/NativeRootStackRenderer.kt
@@ -7,11 +7,17 @@ import androidx.compose.animation.slideInHorizontally
 import androidx.compose.animation.slideOutHorizontally
 import androidx.compose.animation.togetherWith
 import androidx.compose.foundation.background
+import androidx.compose.foundation.isSystemInDarkTheme
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.WindowInsets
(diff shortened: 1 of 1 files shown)


shanerbaner82 (approved):
Custom bottom-bar slot wasn't taking the nav-bar inset, so it drew under the gesture pill. Padding + bleeding the bar color through the inset matches the iOS home-indicator treatment. IME exclusion looks correct.

shanerbaner82:

PR #377, opened 2026-08-25
Merge Info.plist entries structurally so plugins can declare any value type

### LINKED ISSUE #368: iOS plugin plist injection breaks on arrays of dicts (SKAdNetworkItems), blocking any ad-network plugin
`injectPlistEntries()` assumes every plugin plist array is an array of **strings**. Apple has several keys that are arrays of **dicts** — `SKAdNetworkItems` is the one most plugins will hit — and passing one aborts the build with a `TypeError` naming a core class, so it reads as a bug in NativePHP rather than in the plugin.

Any ad-network plugin needs `SKAdNetworkItems`; Apple requires it for attribution. So this currently blocks a whole plugin category on iOS.

## Environment

| | |
|---|---|
| `nativephp/mobile` | `dev-main` @ `bf4fd2a` |
| PHP | 8.4 |
| File | `src/Plugins/Compilers/IOSPluginCompiler.php` |

## What happens

A plugin manifest declaring the shape Apple specifies:

```json
{
  "ios": {
    "info_plist": {
      "SKAdNetworkItems": [
        { "SKAdNetworkIdentifier": "cstr6suwn9.skadnetwork" }
      ]
    }
  }
}
```

`php artisan native:run ios` aborts during "Compiling plugin":

```
TypeError
IOSPluginCompiler::substituteEnvPlaceholders(): Argument #1 ($value)
must be of type string, array given
```

## Cause

`injectPlistEntries()` maps every array element through a string-typed method:

```php
if (is_array($value)) {
    $arrayContent = '';
    foreach ($value as $item) {
        $item = $this->substituteEnvPlaceholders($item);   // <- string $value
        $arrayContent .= "\n\t\t<string>{$item}</string>";  // <- and emits <string>
    }
    ...
```

against

```php
protected function substituteEnvPlaceholders(string $value): string
```

**Both branches fail.** The new-key path above throws, and so does the existing-key path via `mergeArrayEntry()`.

Worth separating two problems, because fixing only the first is not enough:

1. **The type error** — `substituteEnvPlaceholders()` receives an array.
2. **The structure** — even with the type fixed, every element is wrapped in `<string>`, so a dict can never be emitted correctly. Arrays of dicts are not currently representable, not merely unguarded.

## Reproduction

Standalone, no app needed. This extracts `injectPlistEntries()`, `substituteEnvPlaceholders()` and `mergeArrayEntry()` **verbatim** from `main` into a bare class, so it exercises the real code rather than a paraphrase:

```php
<?php
// Harness.php contains the three methods copied unmodified from
// src/Plugins/Compilers/IOSPluginCompiler.php
require __DIR__.'/Harness.php';

$plist = "<?xml version=\"1.0\"?>\n<plist version=\"1.0\">\n<dict>\n</dict>\n</plist>\n";

$entries = ['SKAdNetworkItems' => [
    ['SKAdNetworkIdentifier' => 'cstr6suwn9.skadnetwork'],
]];

$m = new ReflectionMethod($h = new Harness, 'injectPlistEntries');
$m->setAccessible(true);
$m->invoke($h, $plist, $entries);
```

```
TypeError: substituteEnvPlaceholders(): Argument #1 ($value) must be of type string, array given
```

Same input with the key already present in the plist takes `mergeArrayEntry()` and throws identically.

An array of strings works as expected, which is the assumption being encoded:

```php
$m->invoke($h, $plist, ['UIRequiredDeviceCapabilities' => ['armv7', 'metal']]);
// -> <array><string>armv7</string><string>metal</string></array>
```

## Suggested fix

Emit by type rather than assuming string, recursing for nested dicts and arrays, and substituting placeholders only on the strings — something like:

```php
protected function plistValue(mixed $value, int $depth = 2): string
{
    $pad = str_repeat("\t", $depth);

    if (is_array($value)) {
        // A list is <array>, a string-keyed map is <dict>.
        $isList = array_is_list($value);
        $inner = '';

        foreach ($value as $k => $v) {
            if (! $isList) {
                $inner .= "\n{$pad}\t<key>{$k}</key>";
            }
            $inner .= "\n{$pad}\t".$this->plistValue($v, $depth + 1);
        }

        return $isList
            ? "<array>{$inner}\n{$pad}</array>"
            : "<dict>{$inner}\n{$pad}</dict>";
    }

    if (is_bool($value)) {
        return $value ? '<true/>' : '<false/>';
    }

    if (is_int($value)) {
        return "<integer>{$value}</integer>";
    }

    return '<string>'.$this->substituteEnvPlaceholders((string) $value).'</string>';
}
```

That also picks up bools and integers, which hit the same string-typed path today — a manifest declaring `"UIFileSharingEnabled": true` fails for the same reason.

`mergeArrayEntry()` needs the equivalent treatment; merging arrays of dicts also needs an identity to dedupe on, and for `SKAdNetworkItems` that is the inner `SKAdNetworkIdentifier` rather than the element as a whole.

Happy to open a PR if the approach looks right — say which way you would prefer the merge semantics to go and I will follow that.

## Two smaller things noticed nearby

Not the subject of this issue, but they surfaced while tracing it and both cost real debugging time:

- **`injectPlistEntries()` never removes a key.** It adds if missing and updates if present, and it only runs for *registered* plugins. So unregistering a plugin leaves its keys in the generated `Info.plist` indefinitely — an app that has disabled its ads plugin still ships `GADApplicationIdentifier` and `SKAdNetworkItems`. Related: because the values are only re-substituted while the plugin is registered, a changed `.env` value silently does not take effect for an unregistered plugin, leaving a stale id in the built app.
- **A manifest entry whose class is missing is skipped in silence.** Combined with `SomeClass::class` evaluating to a plain string when the class does not exist, a stale provider FQCN in `plugins()` is accepted and quietly does nothing. The visible symptom is "my plugin does not work" with no log line anywhere. A warning in debug would turn that class of problem into a two-minute fix.

gwleuverink:
Fixes #368.

Thanks to @CodyPChristian for the thorough report. A plugin declaring `SKAdNetworkItems` (an array of dicts, required by every ad network) aborted `native:run ios` with a TypeError, because `injectPlistEntries` built the plist by string concatenation and assumed every value is a string or a list of strings. Tracing that assumption turned up more than the crash:

- arrays of dicts threw on both the new-key and existing-key paths
- bools and ints were written as strings; mobile-firebase's `FirebaseAppDelegateProxyEnabled: false` ships as `<string></string>` today and only works because Firebase happens to read an empty string as NO
- merging into a key the base plist already nests (`CFBundleURLTypes`) landed inside the inner `CFBundleURLSchemes` array, since the regex stopped at the first nested `</array>`
- a dict onto `NSAppTransportSecurity`, or a bool onto any existing key, was silently ignored
- a plugin's `resources/ios/Info.plist` lost its bools and arrays, and nested keys were hoisted to the top level
- new string values were not escaped, so `Photos & video` produced a plist Xcode rejects

### The fix

`PlistDocument` reads the plist through the DOM and merges by type: scalars replace, lists union on content, dicts merge key by key, empty arrays and nulls contribute nothing, a type mismatch is replaced by the incoming value. Keys nobody touches keep their original markup byte for byte, and strings are escaped by the DOM.

The compiler opens each plist once and merges every source into it in one pass: the plugin's `resources/ios/Info.plist`, its `info_plist` entries, its background modes, and the app-level overrides last so they still win. The regex helpers are gone. Cody's suggestion to emit by type is the direction taken; the regex merge had to go with it, because once plugins can contribute nested structures its non-greedy `.*?</array>` corrupts them.

This is the second run at the problem. #6 went stale with Shane's review open, and that review doubles as the checklist here: indentation follows depth, bools render as Apple's `<true/>`, empty arrays stay arrays, integers and reals get their own tags, everything is type-hinted, and the tests assert structure by parsing the output back rather than grepping for key names. Placeholders still only resolve in values, not keys, as before.

Two consequences of the union semantics worth knowing. A list never loses items, so a `${PLACEHOLDER}` that changes between builds keeps both values until the scaffold is regenerated, the same way an unregistered plugin's keys already linger. And an app-level override replaces a plugin's scalar but unions with its list.

### Verification

Probe plugin manifest, built on the simulator:

```json
"info_plist": {
    "SKAdNetworkItems": [
        { "SKAdNetworkIdentifier": "cstr6suwn9.skadnetwork" },
        { "SKAdNetworkIdentifier": "${PROBE_NETWORK}" }
    ],
    "FirebaseAppDelegateProxyEnabled": false,
    "NSAppTransportSecurity": { "NSAllowsArbitraryLoads": true },
    "CFBundleURLTypes": [{ "CFBundleURLName": "probe", "CFBundleURLSchemes": ["probe"] }],
    "NSCameraUsageDescription": "Probe: photos & <video>"
}
```

`plutil -p` on the installed bundle's Info.plist:

```
"SKAdNetworkItems" => [
  0 => { "SKAdNetworkIdentifier" => "cstr6suwn9.skadnetwork" }
  1 => { "SKAdNetworkIdentifier" => "4fzdc2evr5.skadnetwork" }
]
"FirebaseAppDelegateProxyEnabled" => 0
"NSAppTransportSecurity" => {
  "NSAllowsArbitraryLoads" => 1
  "NSAllowsArbitraryLoadsInWebContent" => 1
}
"CFBundleURLTypes" => [ 0 => { app entry }, 1 => { "CFBundleURLName" => "probe" ... } ]
"NSCameraUsageDescription" => "Probe: photos & <video>"
```

A second build produced the same plist, so nothing duplicates on rebuild. Unit tests pin the merge rules on `PlistDocument`; compiler tests pin that each source reaches both the device and simulator plists. Suite green.

### Decisions for Simon and Shane

Things the trace surfaced that I left alone because they change behaviour beyond this issue:

- Keys from an unregistered plugin stay in the scaffold plist and ship with the app until `native:install --force`, and a changed `.env` value for such a plugin never lands either. Pre-existing on main. Either track plugin-contributed keys and drop the ones no longer declared, or document `--force` as the way to remove a plugin.
- `compile()` returns early unless a plugin has Swift sources, `info_plist` entries or dependencies, so a plugin whose only iOS data is `background_modes`, `entitlements` or a `resources/ios/Info.plist` file contributes nothing, silently. This PR makes that plist file meaningful for the first time, so the gate bites harder now. Widening it is a few lines; happy to fold it in here.
- A provider class in `plugins()` that matches no installed package is accepted without a word, since `isPluginAllowed` compares strings. A warning at discovery time would turn "my plugin does nothing" into a two-minute fix.
- The entitlements emitters still build XML by hand, keep the first value for a key that already exists, and do not escape. #174 rewrites that path onto its own `PropertyList` codec, so merging both as they stand leaves two plist codecs in the package. Whichever lands second should sit on the other's class; `PlistDocument` can take the entitlements merge as-is if first wins is not something you want to keep.

### DIFF
--- a/src/Plugins/Compilers/IOSPluginCompiler.php
+++ b/src/Plugins/Compilers/IOSPluginCompiler.php
@@ -626,97 +626,34 @@ class IOSPluginCompiler
     }
 
     /**
-     * Merge two plist files
+     * Merge a plugin's Info.plist file into the app's, value types intact.
      */
     protected function mergePlists(string $main, string $plugin): string
     {
-        // Extract key-value pairs from plugin plist
-        preg_match_all('/<key>([^<]+)<\/key>\s*<string>([^<]+)<\/string>/s', $plugin, $matches, PREG_SET_ORDER);
-
-        $entries = [];
-        foreach ($matches as $match) {
-            $entries[$match[1]] = $match[2];
-        }
-
-        return $this->injectPlistEntries($main, $entries);
+        return $this->injectPlistEntries($main, PlistDocument::fromXml($plugin)->all());
     }
 
     /**
-     * Inject entries into plist
+     * Merge entries into the plist, resolving placeholders on the way in.
      */
     protected function injectPlistEntries(string $plist, array $entries): string
     {
-        foreach ($entries as $key => $value) {
-            // Check if key already exists
-            if (str_contains($plist, "<key>{$key}</key>")) {
-                if (is_array($value)) {
-                    $plist = $this->mergeArrayEntry($plist, $key, $value);
-                } elseif (is_string($value)) {
-                    $plist = $this->updateStringEntry($plist, $key, $this->substituteEnvPlaceholders($value));
-                }
+        $document = PlistDocument::fromXml($plist);
+        $document->merge($this->substitutePlaceholdersDeep($entries));
 
-                continue;
-            }
+        return $document->toXml();
+    }
 
-            // Handle array values
-            if (is_array($value)) {
-                $arrayContent = '';
-                foreach ($value as $item) {
-                    $item = $this->substituteEnvPlaceholders($item);
-                    $arrayContent .= "\n\t\t<string>{$item}</string>";
-                }
-                $entry = "\n\t<key>{$key}</key>\n\t<array>{$arrayContent}\n\t</array>";
-            } else {
-                // Handle string values - substitute placeholders
-                $value = $this->substituteEnvPlaceholders($value);
-                $entry = "\n\t<key>{$key}</key>\n\t<string>{$value}</string>";
-            }
-
-            // Add before closing </dict>
-            $plist = preg_replace(
-                '/(\s*<\/dict>\s*<\/plist>)/s',
-                $entry.'$1',
-                $plist,
-                1
-            );
+    /**
+     * Resolve ${ENV_VAR} placeholders in every string of a nested value.
+     */
+    protected function substitutePlaceholdersDeep(mixed $value): mixed
+    {
+        if (is_array($value)) {
+            return array_map(fn ($item) => $this->substitutePlaceholdersDeep($item), $value);
         }
 
-        return $plist;
-    }
-
-    /**
-     * Update an existing string entry's value in the plist
-     */
-    protected function updateStringEntry(string $plist, string $key, string $value): string
-    {
-        $pattern = '/(<key>'.preg_quote($key, '/').'<\/key>\s*<string>)([^<]*)(<\/string>)/';
-
-        return preg_replace_callback($pattern, function ($matches) use ($value) {
-            return $matches[1].htmlspecialchars($value, ENT_XML1 | ENT_QUOTES, 'UTF-8').$matches[3];
-        }, $plist, 1);
-    }
-
-    /**
-     * Merge array values into an existing plist array entry
-     */
-    protected function mergeArrayEntry(string $plist, string $key, array $values): string
-    {
-        $pattern = '/(<key>'.preg_quote($key, '/').'<\/key>\s*<array>)(.*?)(<\/array>)/s';
-
-        return preg_replace_callback($pattern, function ($matches) use ($values) {
-            $existingContent = $matches[2];
-            $newItems = '';
-
-            foreach ($values as $item) {
-                $item = $this->substituteEnvPlaceholders($item);
-                // Only add if not already present
-                if (! str_contains($existingContent, "<string>{$item}</string>")) {
-                    $newItems .= "\n\t\t<string>{$item}</string>";
-                }
-            }
-
-            return $matches[1].$existingContent.$newItems.$matches[3];
-        }, $plist);
+        return is_string($value) ? $this->substituteEnvPlaceholders($value) : $value;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


shanerbaner82:
Landed on main via https://github.com/NativePHP/mobile-air/pull/427, which is this branch rebased over #410 (one conflict, resolved in favour of this PR's structural merge; #410's other changes kept). Your commits are in with original authorship. Thanks for the write-up, and the four follow-up questions at the end are noted.

PR #380, opened 2026-08-25
fix: inline top-bar background never paints on iOS 26.2

SRWieZ:
On iOS 26.2, an inline `<native:top-bar>` with an explicit `background-color` renders as a transparent band: the title and the actions take their colors, but the bar itself never paints. The `toolbarBackground(style)` + `containerBackground(for: .navigation)` pair from #284 only shows the color once content has scrolled under the bar — so a short screen keeps a see-through bar forever.

Repro — any stack screen with:

```blade
<native:top-bar title="Score" background-color="#59390E" text-color="#FAF4EA" display-mode="inline" />
```

| Before | After |
|---|---|
| before | after |

Inline bars have no large-title row to lose to the Liquid Glass visibility bug #284 works around, so they re-request `.toolbarBackground(.visible, for: .navigationBar)`. `large` / `automatic` display modes keep #284's behavior unchanged, and pre-18 already requested visibility. Verified on the iOS 26.2 simulator.

### DIFF
--- a/resources/xcode/NativePHP/NativeRender/NativeRootStackRenderer.swift
+++ b/resources/xcode/NativePHP/NativeRender/NativeRootStackRenderer.swift
@@ -229,7 +229,7 @@ struct NativeRootStackRenderer: View {
             }
             .toolbarColorScheme(toolbarScheme, for: .navigationBar)
             .modifier(HideNavBarModifier(hidden: hideNavBar))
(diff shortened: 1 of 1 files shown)


shanerbaner82 (approved):
Makes sense: large-title on iOS 26 still can't force `.visible` (drops the title row), but inline bars on 26.2 never paint without it. Splitting on `inline` is the right carve-out.

shanerbaner82:

Issue #381, opened 2026-08-25
Password managers (e.g. NordPass) flag app as "unprotected website" because WebView loads over plain HTTP on 127.0.0.1


paul-adams-985:
**Title:** Password managers (e.g. NordPass) flag app as "unprotected website" because WebView loads over plain HTTP on 127.0.0.1

### What were you trying to do?
Log in using NordPass autofill.

### What happened?
NordPass shows an "unprotected website" warning on the login form, because the WebView is served over plain `http://127.0.0.1:<port>` instead of HTTPS. False positive since loopback traffic never leaves the device, but it's alarming to end users. Likely affects other password managers that check origin scheme (1Password, Bitwarden, etc.).

### How to reproduce the bug
1. Build and run any NativePHP app with a login form.
2. Trigger autofill on it with NordPass installed.
3. "Unprotected website" warning appears, since the WebView origin is `http://127.0.0.1:<port>`.

### Debug Output
```
NativePHP Mobile.
+--------------------+--------+
| Package Version    | 4.2.0  |
| PHP Version (Host) | 8.4.23 |
| OS                 | Darwin |
| OS Version         | 25.5.0 |
| Embedded PHP       | 8.4.24 |
+--------------------+--------+

Installed Plugins.
+---------------------------------+---------+
| Package                         | Version |
+---------------------------------+---------+
| codingwithrk/package-info       | v1.1.0  |
| nativephp/mobile-browser        | 1.0.2   |
| nativephp/mobile-firebase       | 1.1.1   |
| nativephp/mobile-network        | 1.0.2   |
| nativephp/mobile-secure-storage | 1.0.2   |
| nativephp/mobile-share          | 1.0.2   |
| nativephp/mobile-ui             | 0.4.0   |
+---------------------------------+---------+
```

### Which operating systems have you seen this occur on?
macOS

### Which platforms were you trying to build for?
iOS (Simulator)

### Notes
Not blocking — autofill still works after dismissing the warning. Would be good if the embedded server could serve over HTTPS (self-signed/local cert) or a custom scheme, so origin-scheme-sensitive tooling treats the WebView as secure.

PR #382, opened 2026-08-25
fix(android): serve the embedded app as https://127.0.0.1, not http

### LINKED ISSUE #148: Android WebView's http://127.0.0.1 origin breaks Google Maps (and any API requiring a secure context) — iOS works, Android doesn't

Render a Google Maps JavaScript map inside the Android WebView (via @googlemaps/js-api-loader). This has worked for months and still works perfectly on iOS — it only fails on Android.

### What happened?

The map fails to render. Google Maps' vector-rendering RPC is blocked by a CORS preflight failure because the response carries no Access-Control-Allow-Origin header for the WebView's http://127.0.0.1 origin (full error in section 4).

  Root cause: Google recently made a server-side change requiring a secure-context origin for its Maps vector RPCs (GetViewportInfo, etc.). The decision is made server-side based on the requesting origin.

  - iOS works because NativePHP serves the WebView over the custom php://127.0.0.1 scheme, which WKWebView treats as a secure context.
  - Android fails because it serves over plain http://127.0.0.1, which is not a secure context.

  This isn't NativePHP-specific — Capacitor/Cordova apps (also http://localhost on Android) hit the identical failure (ionic-team/capacitor discussion #6576).

  We ruled out everything app-side:
  - ❌  API key / referrer restrictions — fails with a fully unrestricted key.
  - ❌ AMaps JS/veesion — pinning v=quarterly makes nofdifferercs (CORS is decided server-side, independent of client JS version).
  - ❌  Rendering type — a raster map still issues the GetViewportInfo RPC.
  - ❌  127.0.0.1 vs localhost — we rewrote the Android origin to http://localhost end-to-end and it still failed. Both are insecure http:// loopback origins.
  
  The only variable that matters is secure context, confirmed by the iOS-vs-Android split. (iOS's php://127.0.0.1 is obviously not on any Google allowlist, so the gate is secure-context, not an origin allowlist.)

  Proposed fix: serve the Android WebView over a secure https:// origin — the Android equivalent of iOS's php:// (custom schemes are treated as insecure/opaque on Android WebView, so it must be https). The standard mechanism is
  WebViewAssetLoader → https://appassets.androidplatform.net, whose underlying behavior is just "load an https:// URL served through shouldInterceptRequest" (no real TLS — secure by scheme), which is the same interception hook
  WebViewManager.shouldInterceptRequest already uses. This would touch the loaded/origin URL and url.contains("127.0.0.1") routing in WebViewManager.kt/MainActivity.kt, the PHP env (HTTP_HOST/APP_URL/ASSET_URL in LaravelEnvironment.kt, PHP.c,
  php_bridge.c), and the cookie/session domain (plus SESSION_SECURE_COOKIE=true since it's now https).

### How to reproduce the bug

1. Build any NativePHP Mobile app for Android that loads a Google Maps JS vector map (@googlemaps/js-api-loader → importLibrary('maps') → new google.maps.Map(...)).
2. Run it on an Android emulator/device (php artisan native:run android).
3. Open the screen containing the map.
4. The map fails to render. Inspect via chrome://inspect → the WebView's DevTools → Network/Console to see the CORS preflight failure on the GetViewportInfo request.
5. Build the same app for iOS (php artisan native:run ios) → the map renders correctly. The only difference is the WebView origin (http://127.0.0.1 on Android vs the secure php://127.0.0.1 on iOS).

### Debug Output

 Access to fetch at
  'https://maps.googleapis.com/$rpc/google.internal.maps.mapsjs.v1.MapsJsInternalService/GetViewportInfo'
  from origin 'http://127.0.0.1' has been blocked by CORS policy:
  Response to preflight request doesn't pass access control check:
  No 'Access-Control-Allow-Origin' header is present on the requested resource.

  - WebView origin (Android, failing): http://127.0.0.1
  - WebView origin (iOS, working): php://127.0.0.1

macOS

iOS (Simulator), Android (Device)

### Notes

_No response_

paul-adams-985:
Fixes #148 (secure-context break) and the Android side of #381 (password-manager "unprotected site" warning).

## What

Android's WebView loads the embedded app at `http://127.0.0.1`, a genuinely insecure origin. iOS avoids this with a `php://` scheme handler; Android had no equivalent, so anything gated on a secure context (Maps JS SDK, password-manager autofill checks, `isSecureContext`) treats the app as insecure.

Every `127.0.0.1` request is already fully intercepted in `shouldInterceptRequest` and dispatched in-process to the embedded PHP runtime by path alone — no real socket, no TLS handshake. So flipping the scheme to `https` needs no new networking layer (no `WebViewAssetLoader`, no cert, no new dependency) — just consistent URL construction across the existing call sites. 12 lines, 5 files.

Also updates `WebCookieMirror`/`WebRenderer` to mirror cookies against the new `https://127.0.0.1` origin — otherwise `Secure`-flagged cookies silently fail to write (same root cause as #334; that PR's cookie-scheme patch becomes redundant once this lands, cc @inoha-kudo).

## What this doesn't touch

- iOS — already unaffected (untouched here).
- The Jump remote-dev-forward path (`WebViewManager`'s `JumpWebViewSession` branch, `PHPWebViewClient.forwardToRemote`) — left on `http` deliberately, dev-only tooling, out of scope.
- Laravel's own `$request->secure()` (no `HTTPS`/`X-Forwarded-Proto` server var threaded through) — `APP_URL`/`ASSET_URL` cover URL generation, but framework-level secure-request detection may still need a follow-up if anything depends on it.

## Testing

**I have not been able to test this on a device or emulator** — no Android environment available where this was written. The change is a straightforward scheme substitution over an already-fully-intercepted request path, but this needs real-device verification before merge:

- [ ] Cold boot reaches login/home without errors
- [ ] Login/session/CSRF flow works end-to-end (the exact path #334 found breaking)
- [ ] Google Maps SDK / other secure-context-gated API from #148 now works
- [ ] Password manager no longer shows "unprotected website"
- [ ] Jump dev/hot-reload flow still works unaffected

Opening as a draft for that reason — happy to iterate based on device testing from anyone who can run it, or reviewer feedback on the approach itself.

### DIFF
--- a/resources/androidstudio/app/src/main/java/com/nativephp/mobile/bridge/LaravelEnvironment.kt
+++ b/resources/androidstudio/app/src/main/java/com/nativephp/mobile/bridge/LaravelEnvironment.kt
@@ -842,8 +842,10 @@ class LaravelEnvironment(private val context: Context) {
 
             setEnvironmentVariables(
                 // Laravel environment settings
-                "APP_URL" to "http://127.0.0.1",
-                "ASSET_URL" to "http://127.0.0.1/_assets",
+                // https: 127.0.0.1 must be a secure context (#148); fully
+                // intercepted below, so generated URLs never hit real TLS.
+                "APP_URL" to "https://127.0.0.1",
+                "ASSET_URL" to "https://127.0.0.1/_assets",
                 "DB_CONNECTION" to "sqlite",
                 "DB_DATABASE" to "${appStorageDir.absolutePath}/persisted_data/database/database.sqlite",
                 "CACHE_DRIVER" to "file",
--- a/resources/androidstudio/app/src/main/java/com/nativephp/mobile/network/WebRenderer.kt
+++ b/resources/androidstudio/app/src/main/java/com/nativephp/mobile/network/WebRenderer.kt
@@ -43,7 +43,7 @@ class WebRenderer(
         WebCookieMirror.markWebViewAvailable()
         val jar = CookieManager.getInstance()
         LaravelCookieStore.all().forEach { (name, value) ->
-            jar.setCookie("http://127.0.0.1", "$name=$value")
+            jar.setCookie("https://127.0.0.1", "$name=$value")
         }
         jar.flush()
 
--- a/resources/androidstudio/app/src/main/java/com/nativephp/mobile/network/WebViewManager.kt
+++ b/resources/androidstudio/app/src/main/java/com/nativephp/mobile/network/WebViewManager.kt
@@ -248,7 +248,7 @@ class WebViewManager(
                 // Handle relative URLs (convert to php://)
                 if (url.startsWith("/")) {
                     val uri = request.url
-                    val fullUrl = "http://127.0.0.1${uri.encodedPath}" +
+                    val fullUrl = "https://127.0.0.1${uri.encodedPath}" +
                         (uri.encodedQuery?.let { "?$it" } ?: "")
 
                     Log.d(TAG, "🛠️ Rewriting relative URL with query: $fullUrl")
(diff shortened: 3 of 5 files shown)


paul-adams-985:
Closing for now — want to get this verified on a real device before it's in front of reviewers. Will reopen (or open fresh) once that's done.

PR #395, opened 2026-08-26
fix: Android bottom-bar draws under the system navigation bar

SRWieZ:
> Replacement for #375, which was accidentally merged and then reverted. Reopened from the original branch at the maintainer’s request.

## What's wrong
On Android, `<native:bottom-bar>` content draws **underneath the system navigation bar** (gesture pill or 3-button strip). Material's own `NavigationBar` pads itself above the system bar, but the stack renderer drops custom bottom-bar content into the Scaffold slot raw — no insets — so it sits flush with the window edge, and the home button lands on top of your buttons.

Minimum repro:

```blade
<native:bottom-bar>
    <native:row class="w-full items-center justify-between px-5 py-3 bg-white">
        <native:column class="w-12 h-12 rounded-full bg-zinc-200 items-center justify-center"><native:text font="bold" class="font-bold">1</native:text></native:column>
        <native:pressable class="px-8 h-[44] rounded-full bg-blue-600 items-center justify-center"><native:text font="semibold" class="font-semibold text-white">Primary</native:text></native:pressable>
        <native:column class="w-12 h-12 rounded-full bg-zinc-200 items-center justify-center"><native:text font="bold" class="font-bold">2</native:text></native:column>
    </native:row>
</native:bottom-bar>
```

## What this does
- Pads the bottom-bar slot with the navigation-bar inset and paints the bar's own background color through the inset — the same "bleed" treatment the iOS renderer gives the home-indicator strip, so the strip matches the bar instead of showing the window background.
- The IME inset is excluded: the root `imePadding` already lifts the bar above the keyboard, so keeping the nav-bar inset there would float it too high.
- The tabs renderer needs no change (its bottom bar already sits inside the Scaffold padding). No iOS or PHP changes.

## Before / after (Android emulator, API 36, 3-button navigation, same Blade)
Before: the bar is flush with the window edge — the home button sits ON the Primary button. After: the bar rides above the nav strip and its background bleeds through it.

### DIFF
--- a/resources/androidstudio/app/src/main/java/com/nativephp/mobile/ui/nativerender/NativeRootStackRenderer.kt
+++ b/resources/androidstudio/app/src/main/java/com/nativephp/mobile/ui/nativerender/NativeRootStackRenderer.kt
@@ -7,11 +7,17 @@ import androidx.compose.animation.slideInHorizontally
 import androidx.compose.animation.slideOutHorizontally
 import androidx.compose.animation.togetherWith
 import androidx.compose.foundation.background
+import androidx.compose.foundation.isSystemInDarkTheme
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.WindowInsets
 import androidx.compose.foundation.layout.WindowInsetsSides
+import androidx.compose.foundation.layout.exclude
+import androidx.compose.foundation.layout.ime
+import androidx.compose.foundation.layout.navigationBars
+import androidx.compose.foundation.layout.windowInsetsPadding
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.only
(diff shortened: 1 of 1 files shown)


shanerbaner82 (approved):
Reviewed against the surrounding Kotlin rather than the description.

The bleed colour resolution mirrors the established path exactly:

```kotlin
// this PR
val darkBg = if (isSystemInDarkTheme()) inner.props.getColor("dark_bg_color", 0) else 0
val barBg  = if (darkBg != 0) darkBg else (inner.style?.bgColor ?: 0)

// NodeModifiers.kt:158
val bgArgb = if (darkBg != 0) darkBg else style.bgColor
```

All four helpers it leans on already exist (`getColor`, `argbToComposeColor`, `style.bgColor`, `dark_bg_color`), and `argbToComposeColor` is already in this file. Byte-identical to #375.

On why a `safe-area-bottom` class isn't the answer here: it would mostly work today — Android's safe area is a root CompositionLocal (`LocalSafeAreaBottom`, from `ViewCompat.getRootWindowInsets(...).systemBars().bottom`), not a Scaffold-consumed inset, so it is non-zero inside the bar slot, and `background` precedes `padding` in `NodeModifiers` so the bleed would happen too. But it would diverge from iOS, where `StackBottomBarInsetModifier` already does this automatically, and it reads `systemBars()`, which knows nothing about the root `imePadding()` — keyboard up, the bar would be lifted by the keyboard *and* padded by the nav bar. The `.exclude(WindowInsets.ime)` here is what avoids that.

Three non-blocking edge cases where the bleed colour can differ from what the bar paints:

1. Gradient bars — `NodeModifiers` prefers a gradient over the flat colour; this reads `bgColor` only, so a `bg-gradient-*` bar bleeds the flat colour (often nothing).
2. Elevation fallback — `NodeModifiers` paints `Color.White` when elevation > 0 and no bg is set; nothing bleeds here.
3. Explicit transparent — `NodeModifiers` skips `Color.Transparent`; this checks `!= 0` only. Harmless, painting transparent is a no-op.

None regress anything: today there is no bleed and the bar sits under the nav bar. iOS's `barBackgroundColor()` is flat-colour-only too, so this preserves parity rather than breaking it.

Caveat on this review: it is a code read, not a device run — CI covers no Kotlin. The before/after on API 36 with 3-button nav is the real evidence. The one case I could not verify is the keyboard path (`.exclude(WindowInsets.ime)`): a bar containing a text input, keyboard up, should not float a nav-bar height too high.

PR #400, opened 2026-08-27
Apply pending OTA zips on boot; take check/download out of core

shanerbaner82:
OTA check and download never belonged in the runtime. They talk to Bifrost, they care about zip names and UX (manual / prompt / silent), and that is all plugin work. Core should do one thing on boot: if a pending zip is sitting there, extract it instead of (or after) the bundled payload, and do not wipe the app's `.env` when you do.

That is this PR. iOS already applied `Documents/updates/*.zip` on next launch; Android now does the same with `{appStorageDir}/updates/*.zip`. Both stash `.env` around the extract. The in-core Bifrost client is gone on both platforms. DEBUG still re-extracts the bundle every boot so `native:run` shows local PHP, and it now skips leftover pending zips, otherwise a queued update covers your working copy. Versioned store builds still apply pending on the next cold start. Bundled names stay `app.zip` / `laravel_bundle.zip` (renaming those is a packager lockstep break). A pending zip is additive, so existing installs are fine. The plugin that drops `pending.zip` will need a min core version that understands this.

Five commits, in order: preserve `.env` on iOS apply, Android next-boot apply, rip the iOS client, rip the Android client, skip pending in DEBUG.


simonhamp (approved):
Looks good. Let's go

PR #401, opened 2026-08-28
Split the query string off a native route URI before resolving it

### LINKED ISSUE #290: Warm deep links with query parameters are not resolved or passed to NativeComponent data
I am implementing a browser-based authentication flow in a SuperNative application.

The app opens an authentication page using `Browser::auth()`. After a successful login, the authentication server redirects to the app using a custom URL scheme containing a one-time authorization code:

```text
myapp://auth/callback?authCode=example
```

The callback screen is registered as a native route:

```php
Route::native('/auth/callback', AuthCallback::class);
```

I expect the callback component to receive the query parameter as navigation data:

```php
$authCode = $this->data('authCode');
```

Environment:

- NativePHP Mobile: `4.0.1`
- NativePHP Browser: `1.0.2`
- Laravel: `13`
- PHP: `8.5`
- UI runtime: SuperNative / EDGE

### What happened?

When the application is already running and receives the callback as a warm deep link, the query parameter is unavailable in the callback component.

The deep link is normalized correctly:

```text
/auth/callback?authCode=example
```

However, the component does not receive the value:

```php
$this->data('authCode'); // null
```

Using the current request as a fallback does not work either:

```php
request()->query('authCode'); // null
```

I expected NativePHP to:

1. Resolve the route using only `/auth/callback`.
2. Parse `authCode=example` from the query string.
3. Pass the parsed query parameters to the component as navigation data.

Expected:

```php
$this->data('authCode') === 'example';
```

Actual:

```php
$this->data('authCode') === null;
```

The issue appears to occur in the SuperNative warm deep-link navigation path.

`DeepLinkRouter` preserves the query string and sends the complete URI through the `__deeplink` event. `NativeComponent` then creates a `NavigationIntent` with that URI:

```php
new NavigationIntent(
    NavigationIntent::NAVIGATE,
    $uri,
);
```

`NativeRouter::resolve()` compares the complete URI against the registered route:

```text
Registered route: /auth/callback
Received URI:     /auth/callback?authCode=example
```

`NativeRouter` does not separate the path and query string before route resolution, and it does not convert the query parameters into navigation data.

This appears to be a regression from v3. In v3, the deep-link URL was loaded directly through the WebView with its query string preserved, so Laravel could access the value using `request()->query()`.

### How to reproduce the bug

1. Configure a custom deep-link scheme:

```dotenv
NATIVEPHP_DEEPLINK_SCHEME=myapp
```

2. Register a native callback route:

```php
use App\NativeComponents\Auth\AuthCallback;
use Illuminate\Support\Facades\Route;

Route::native('/auth/callback', AuthCallback::class);
```

3. Create the callback component:

```php
<?php

namespace App\NativeComponents\Auth;

use Illuminate\View\View;
use Native\Mobile\Edge\NativeComponent;

class AuthCallback extends NativeComponent
{
    public ?string $receivedAuthCode = null;

    public function mount(): void
    {
        $this->receivedAuthCode = $this->data('authCode');
    }

    public function render(): View
    {
        return view('native.auth.auth-callback');
    }
}
```

4. Open an authentication page from an active native screen:

```php
use Native\Mobile\Facades\Browser;

public function login(): void
{
    Browser::auth('https://example.com/login');
}
```

5. Make the authentication page redirect to:

```text
myapp://auth/callback?authCode=example
```

6. Start the app and leave a native screen active.

7. Trigger the login action and complete authentication.

8. Allow the browser to redirect to the running app.

9. Inspect the callback component.

Actual result:

```php
$this->data('authCode') === null;
```

Expected result:

```php
$this->data('authCode') === 'example';
```

The same issue can be reproduced with a generic warm deep link:

```text
myapp://auth/callback?foo=bar
```

Expected:

```php
$this->data('foo') === 'bar';
```

### debug output

   INFO  NativePHP Mobile.

+--------------------+--------+
| Package Version    | 4.0.1  |
| PHP Version (Host) | 8.5.8  |
| OS                 | Darwin |
| OS Version         | 24.6.0 |
| Embedded PHP       | 8.5.8  |
+--------------------+--------+

   INFO  Installed Plugins.

+---------------------------------+---------+
| Package                         | Version |
+---------------------------------+---------+
| nativephp/mobile-browser        | 1.0.2   |
| nativephp/mobile-secure-storage | 1.0.2   |
| nativephp/mobile-ui             | 0.3.0   |
+---------------------------------+---------+

   INFO  Development Tools.

| Xcode          | Xcode 16.4                           |
| Android Studio | Android Studio AndroidStudio2025.1.2 |
| Java           | openjdk version "24.0.2" 2025-07-15  |

simonhamp:
NativeRouter::resolve() matched the whole URI against patterns registered as paths, so a deep link with a query string could not find its screen. The parsed query now reaches the screen as `$this->data()`.

Fixes #290.

### DIFF
--- a/src/Edge/NativeRouter.php
+++ b/src/Edge/NativeRouter.php
@@ -174,25 +174,60 @@ class NativeRouter
         static::$currentGroupLayout = null;
     }
 
-    public static function resolve(string $uri): ?array
+    /**
+     * Split a navigation URI into its path and its decoded query parameters.
+     *
+     * Route patterns are registered as paths, so anything after `?` has to come
+     * off before matching or `/auth/callback?code=x` misses `/auth/callback`.
+     * `parse_str` is what gives array syntax (`tag[]=a&tag[]=b`, the shape a
+     * Livewire array `#[Url]` prop serialises to) the same meaning it has in a
+     * web request.
+     *
+     * @return array{0: string, 1: array<string, mixed>}
+     */
+    protected static function splitQuery(string $uri): array
     {
         $uri = '/'.ltrim($uri, '/');
 
-        // Match on the path alone. A query string would otherwise be swallowed
-        // by the last route parameter — "/docs/{page}?utm_source=x" matched with
-        // page = "installation?utm_source=x", so a tagged link resolved to a
-        // screen that then found nothing. Tagged links are the common case for
-        // anything shared publicly.
-        $uri = explode('?', $uri, 2)[0];
+        // A fragment addresses somewhere within a document. It is never part of
+        // a route or its parameters, and the hosts pass it through untouched.
+        if (($hash = strpos($uri, '#')) !== false) {
+            $uri = substr($uri, 0, $hash);
+        }
+
+        if (($mark = strpos($uri, '?')) === false) {
+            return [$uri, []];
+        }
+
+        parse_str(substr($uri, $mark + 1), $query);
+
+        // `?a=1` on its own leaves an empty path, and no route is registered
+        // under ''. Root is what a bare query string addresses.
+        $path = substr($uri, 0, $mark);
+
+        return [$path === '' ? '/' : $path, $query];
+    }
+
+    /**
+     * Resolve a navigation URI to the screen registered for it.
+     *
+     * `query` is the URI's query string, parsed. Callers merge it into the
+     * component's navigation data, so a deep link's parameters reach the screen
+     * through `$this->data()` the same way an in-app `navigate()` payload does.
+     */
+    public static function resolve(string $uri): ?array
+    {
+        [$path, $query] = static::splitQuery($uri);
 
         // Exact match first
-        if (isset(static::$routes[$uri])) {
-            $entry = static::$routes[$uri];
+        if (isset(static::$routes[$path])) {
+            $entry = static::$routes[$path];
 
             return [
                 'class' => $entry['class'],
                 'layout' => $entry['layout'] ?? null,
                 'params' => [],
+                'query' => $query,
             ];
         }
 
@@ -201,13 +236,14 @@ class NativeRouter
             $regex = preg_replace('/\{(\w+)\}/', '(?P<$1>[^/]+)', $pattern);
             $regex = '#^'.$regex.'$#';
 
-            if (preg_match($regex, $uri, $matches)) {
+            if (preg_match($regex, $path, $matches)) {
                 $params = array_filter($matches, fn ($key) => is_string($key), ARRAY_FILTER_USE_KEY);
 
                 return [
                     'class' => $entry['class'],
                     'layout' => $entry['layout'] ?? null,
                     'params' => $params,
+                    'query' => $query,
                 ];
             }
         }
@@ -353,6 +389,7 @@ class NativeRouter
                 $component = $this->createComponent(
                     $resolved['class'],
                     $resolved['params'] ?: ($entry['params'] ?? []),
+                    $resolved['query'],
                 );
                 if (! empty($resolved['layout'])) {
                     $component->setLayout($resolved['layout']);
@@ -378,9 +415,13 @@ class NativeRouter
      * Entry point. Init shared memory, run the navigation loop,
      * shutdown when done.
      *
+     * `$data` is the launch screen's navigation data — on a cold deep link the
+     * request's query parameters, so `$this->data()` answers the same on the
+     * first screen as it does on every screen navigated to after it.
+     *
      * @return string|null Exit URI for redirect, or null
      */
-    public function start(string $class, array $params = [], string $uri = ''): ?string
+    public function start(string $class, array $params = [], string $uri = '', array $data = []): ?string
     {
         NativeComponent::registerDumpHandler();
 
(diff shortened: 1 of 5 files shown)


shanerbaner82 (commented):
Full suite passes at the head (961 passed), so everything below is something the tests don't cover. Verified each with a `php -r` / Pest reproduction unless noted. Most severe first.

**1. Android decodes the query twice.** `MainActivity.kt:704` (and :666 for the FCM path, plus the cold `loadUrl()`) builds the warm `__deeplink` URI from `uri.query`, which is already decoded; the new `parse_str` in `NativeRouter::splitQuery` decodes again. iOS forwards the still-encoded `url.query`. `myapp://auth/callback?code=a%2Bb%3D%3D&state=x%26y` gives Android `code='a b=='`, `state='x'` plus a bogus key `y`; iOS gives `code='a+b=='`, `state='x&y'`. The "decodes a percent-encoded value exactly once" test only holds for iOS input. Fix: `uri.encodedQuery`, which every other Android forwarding site already uses.

**2. `parse_str` is a new throw site.** `NativeRouter.php:202`: a query over `max_input_vars` (1000, no override in either embed) raises E_WARNING, which HandleExceptions turns into an ErrorException. A warm deep link sets the NAVIGATE intent, `loop()` calls `resolve()` at :521 outside its try, the exception reaches `start()`'s catch at :448 which returns null, and the native runloop ends. A verified app link takes the EVENT_NATIVE path instead and shows the error screen. Reproduced: 1001 pairs → "Input variables exceeded 1000".

**3. Stack URIs keep the query.** `NativeRouter.php:549`: NAVIGATE/REPLACE push `'uri' => $intent->uri` verbatim, and `currentUri()` feeds that to `TabBar::highlight()` (`TabBar.php:222`, exact / prefix compare) and the host's `currentUri` prop. `/home?ref=push` highlights no tab, and `NativeRootTabsRenderer.swift:377` gets a nil owning tab and caches it as a distinct screen. Pre-PR that URI resolved null and exited to web, so this is newly reachable. Cold links store the stripped path, so the same link stores different URIs cold vs warm.

**4. Cold and warm disagree on data.** `NativeServiceProvider.php:384` reads `request()->query->all()` after TrimStrings and ConvertEmptyStringsToNull have run; warm links use raw `parse_str`. `/auth/callback?code=&state=%20abc%20` cold: `data('code') === null`, `data('state') === 'abc'`; warm: `''` and `' abc '`. So `data('code', 'missing')` answers differently, against the new `start()` docblock's promise.

**5. Arrays reach `data()` unguarded.** `myapp://detail/7?from[]=a` → `['from' => ['a']]`; the PR's own `DetailScreen` fixture throws "Array to string conversion" (reproduced via `Native::visit('/detail/9?from[]=a')` and via the `__deeplink` event). On device that's the error screen for a link anyone can craft. `?user.name=x` lands under `user_name`, so `data('user.name')` is silently null.

**6. Trust boundary.** Untrusted deep-link params now merge into the same bag as trusted in-app `navigate()` payloads (`:540`), and the upgrade doc describes `data()` as the in-app mechanism. Custom schemes are unverified, so `myapp://checkout?verified=1&user_id=42` steers any screen doing `if ($this->data('verified'))`. At minimum an upgrade note; arguably a separate `query()` accessor.

**7. Hot restart never matches a query-bearing top screen.** `NativeServiceProvider.php:361` compares the saved top URI (stored verbatim with its query) against a query-stripped `request()->path()`. On `/detail/7?from=x`, save a file: :361 returns a redirect before `start()` runs, `PHPBridge.kt:458` / `HotReloadServer.swift:226` discard the response, the app sits on a dead frame with `.hot_restart` left on disk. The new "restores query data onto a hot-reloaded stack" test calls `preloadStack()` directly and never reaches this compare. Dev loop only.

**8. Hot reload restores query data only.** `start()` records the stripped path and never stores `$data`; `getStackEntries()` serialises uri+params; `preloadStack()` rebuilds from the query alone. So `navigate('/detail/7?from=query', ['from' => 'payload'])` renders "payload" live and "query" after reload, inverting the payload-over-query rule this PR introduces at :540, and `navigate('/detail/3', ['from' => 'cart'])` loses `from` entirely. Dev loop only.

**9. The merge is hand-copied at five sites** (`NativeRouter` :392/:540/:596, `TestableComponent` :160/:622) and `start()` is a sixth `resolve()` caller that doesn't merge at all (creates the component at :432, resolves at :436, ignores `$resolved['query']`, relies on the provider's 4th `$data` argument). Verified: `start(CounterScreen::class, [], '/auth/callback?authCode=x')` gives `data('authCode') === null` despite the docblock. Any second `start()` caller re-creates #290.

**10. Fragment stripping rests on a false comment.** `NativeRouter.php:192` says the hosts pass fragments through; `DeepLinkRouter.swift:136` and `MainActivity.kt:665-719` never read them. The branch makes `navigate('#top')` push a duplicate root with `'uri' => '#top'`, and `JumpCommand.php:482/1126` now calls `/settings#account` native while `BootPlanner.swift:55` / `BootPlanner.kt:41` (split on `?` only) pick the WebView boot.

One root fix covers 3, 7, 8 and most of 9: store the split path (and the merged data) in the stack entry instead of the verbatim `navigate()` string, and derive the query once inside `NativeRouter` rather than at five call sites plus a `start()` parameter. I'm putting that together as a PR into this branch so it's concrete rather than a wish list; 1 is a one-token change in the Android shell I'll include.

PR #407, opened 2026-08-31
fix: make native:check-build-number work for iOS

CodyPChristian:
The iOS check could never succeed. It always reported `Could not check App Store Connect` and then suggested build number `1`.

```
$ php artisan native:check-build-number ios --api-key=/path/to/AuthKey.p8

 Could not check App Store Connect: The "api-key-path" option does not exist.
📱 App Store: No releases found or check not implemented
💻 Local current: 2
💡 Suggested next: 1
```

## Cause

`ChecksLatestBuildNumber` is shared by three commands that declare different subsets of the same flags, and it read four of them unconditionally — `api-key-path`, `api-key-id`, `api-issuer-id` and `google-service-key`:

```php
$apiKeyPath = $this->option('api-key-path') ?? $this->option('api-key') ?? env('APP_STORE_API_KEY_PATH');
```

Illuminate's `option()` **throws** `InvalidArgumentException` for an option the command has not defined, rather than returning null — so the throw happens before `??` can fall through to the env fallback. `BuildIosAppCommand` and `PackageCommand` declare `--api-key-path` and `--api-key-id`, so they are unaffected. `CheckBuildNumberCommand` declares only `--api-key`, so every lookup threw and was swallowed by the surrounding `catch`, surfacing as the generic warning above.

Fixing only `api-key-path` is not enough — with that alone the next call throws on `api-key-id` instead. (Found the hard way.)

These reads now go through a small `consoleOption()` helper that checks the command's own definition first, so each command contributes whichever flags it declares and the env fallback still applies to the rest. No command's CLI surface changes.

## Also: the fallback suggestion

That branch is reached when the lookup **fails**, not only when an app is genuinely new, and it suggested `1` unconditionally. For any app that has already shipped, `1` is a build number App Store Connect has long since seen and will reject on upload — so following the suggestion costs a full build and upload cycle to discover.

It now suggests one past the local value when there is one. The Android branch had the same flat `1` and is fixed alongside it.

## Verified against a live App Store Connect account

After:

```
  Builds found ............................. 5 for version 1.0
  Latest build ............................... 7 (version 1.0)
📱 App Store latest: 7
💻 Local current: 2
💡 Suggested next: 8
```

Two files, no behaviour change for the commands that already worked.

### DIFF
--- a/src/Commands/CheckBuildNumberCommand.php
+++ b/src/Commands/CheckBuildNumberCommand.php
@@ -78,7 +78,9 @@ class CheckBuildNumberCommand extends Command
                 $this->line('🔧 To update: add --update flag');
             }
         } else {
-            $baseSuggested = 1;
+            // Same reasoning as the iOS branch: this is also reached when the
+            // Play Store lookup fails, not only for a genuinely new app.
+            $baseSuggested = $currentLocal ? ((int) $currentLocal + 1) : 1;
             $suggested = $baseSuggested + $jumpBy;
             $this->line('📱 Play Store: No releases found (new app)');
             $this->line('💻 Local current: '.($currentLocal ?: 'not set'));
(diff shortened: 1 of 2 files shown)


simonhamp:
Looks to be a duplicate of #293.

Issue #10, opened 2026-02-08
Android build fails: PHP binaries for NativePHP Mobile are missing / failed to download


ChristYoga123:
### Description

Building NativePHP v3 for Android consistently fails due to missing PHP Android
binaries (`libphp.so`). NativePHP attempts to download Android PHP binaries,
but the process ends with a failure and the build continues, causing CMake/Ninja
to fail later.

---

### PHP binaries download

Downloading Android PHP binaries ........................................ DONE
Failed to download PHP binaries.

---

### CMake configuration (works as expected)

Task :app:configureCMakeDebug

Checking the license for package CMake 3.22.1 in
/Users/mac/Library/Android/sdk/licenses
License for package CMake 3.22.1 accepted.
Preparing "Install CMake 3.22.1 v.3.22.1".
Installing CMake 3.22.1 in
/Users/mac/Library/Android/sdk/cmake/3.22.1
"Install CMake 3.22.1 v.3.22.1" complete.

This indicates that Android SDK and CMake are correctly installed and detected.

---

### Build failure

Task :app:buildCMakeDebug[compat,php_wrapper] FAILED

ninja: error:
'/android/app/src/main/jniLibs/arm64-v8a/libphp.so',
needed by 'libphp_wrapper.so',
missing and no known rule to make it

Execution failed for task ':app:buildCMakeDebug[compat,php_wrapper]'.

C++ build system [build] failed while executing:
/Users/mac/Library/Android/sdk/cmake/3.22.1/bin/ninja

---

### Expected behavior

One of the following:
1. Official prebuilt PHP binaries for Android (arm64-v8a) are provided
2. Clear documentation on how to build PHP for Android manually
3. Android PHP wrapper build is skipped automatically if PHP binaries are missing
4. Android support is clearly marked as experimental / incomplete

---

### Actual behavior

- Android PHP binaries download step runs but fails
- Build process continues
- CMake/Ninja fails because `libphp.so` does not exist
- Build fails even though Android SDK and CMake are correctly installed

---

### Environment

- NativePHP version: v3
- Platform: Android
- ABI: arm64-v8a
- Host OS: macOS
- Android SDK path: `/Users/mac/Library/Android/sdk`
- CMake: 3.22.1 (installed automatically)
- Build system: Gradle + CMake + Ninja

---

### Additional notes

- `sdk.dir` is not explicitly set in `local.properties`, but Android SDK is
  correctly detected and used by Gradle (CMake installs successfully).
[...]

PR #409, opened 2026-08-31
fix(ios): prevent silent corruption/truncation of binary PHP response bodies

sonalidudhia:
## Problem

The native bridge returns the whole HTTP response (headers + body) from C to Swift as a single NUL-terminated C string. Binary bodies do not survive that trip:

- `do_dispatch()` and `do_webview_dispatch()` in `PHP.c` hand the response to Swift via `strdup()`, which relies on `strlen()` — any embedded NUL byte (common in JPEG/PNG/PDF) truncates the response silently.
- `PersistentPHPRuntime.dispatch()` / `WebviewPHPRuntime.dispatch()` read it back with `String(cString:)`, mangling non-UTF-8 bytes.

In practice the response usually dies even earlier: `PHPSchemeHandler.forwardToPHP()` does `String(data:encoding:.utf8)` over the entire response, which returns `nil` for a binary body. The user sees a broken image with no error anywhere.

This is the iOS manifestation of the same bridge bug as the Android crash in #10 (there it is `JNI NewStringUTF` on non-UTF-8 bytes; here it fails silently instead).

## Fix

Capture the body with output buffering around `sendContent()` in both dispatch paths, and base64-encode it when it is not valid UTF-8 **or contains a NUL byte**, flagged with an internal `X-Native-Body-Encoding: base64` header. Base64 is pure ASCII with no NULs, so it survives `strdup`/`strlen` and `String(cString:)` intact.

`PHPSchemeHandler.forwardToPHP()` detects the header, strips it before it reaches the WebView, and base64-decodes the body back to raw `Data` (with `.ignoreUnknownCharacters`, so stray whitespace cannot turn a valid response into a 500).

Notes:
- `sendContent()` (not `getContent()`) still generates the body, so `StreamedResponse` / `BinaryFileResponse` keep working. This is not true streaming either way — `capture_php_output()` already accumulates the whole response in a thread-local buffer before handoff — so `ob_start()` adds one in-memory copy plus base64's 33% overhead, on binary bodies only.
- Text/JSON/HTML responses are byte-for-byte unchanged: no marker header, no encoding.
- The eval'd PHP is kept comment-free and `eval_code` grown 8K → 16K. That string is re-parsed by `zend_eval_string` on every request and shares a fixed buffer with the substituted request URI, where an overflow would be silently truncated by `snprintf` into malformed PHP.
- The legacy "classic mode" fallback (`NativePHPApp.laravel(request:)`) is intentionally untouched — it needs `bootstrap/ios/native.php`, outside this change.

## Verification

Tested on iPhone 17 Pro simulator (iOS 26) with a route returning `response()->file()` on a 26,015-byte JPEG.

| Build | Result |
|---|---|
| Unpatched bridge | broken-image placeholder |
| Patched bridge | image renders correctly |

Bridge-level instrumentation on the patched build:

```
status=200 isBase64=true bodyStringCount=34688
headers=["content-type": "image/jpeg", "content-length": "26015", ...]
```

`34688` is exactly base64 of 26,015 bytes, confirming the full body crossed the boundary; the marker header is absent from the dict handed to the WebView, confirming it is stripped.

Also verified: both generated PHP blocks pass `php -l`; encode/decode round-trips byte-identically for binary, UTF-8 text, HTML, and valid-UTF-8-containing-NUL bodies; normal HTML/JSON pages and Laravel's error page render unchanged through the patched bridge.

### DIFF
--- a/resources/xcode/Include/Bridge/PHP.c
+++ b/resources/xcode/Include/Bridge/PHP.c
@@ -456,8 +456,24 @@ static void do_dispatch(const dispatch_params_t *params) {
         SG(request_info).content_length = 0;
     }
 
-    // Build dispatch eval code — same pattern as Android
-    static char eval_code[8192];
+    // Build dispatch eval code — same pattern as Android.
+    //
+    // The body is captured with ob_start()/ob_get_clean() around sendContent()
+    // rather than echoed straight through, so it can be inspected before it
+    // crosses the C -> Swift boundary. sendContent() (not getContent()) is still
+    // what generates it, so StreamedResponse / BinaryFileResponse keep working.
+    //
+    // The whole HTTP response (headers + body) travels back to Swift as one
+    // NUL-terminated C string (strdup'd below, then String(cString:) on the
+    // Swift side). A binary body (image, PDF, etc.) breaks that two ways: an
+    // embedded NUL byte truncates the response outright via strdup/strlen, and
+    // non-UTF-8 bytes get mangled by String(cString:). So the body is base64-
+    // encoded whenever it isn't valid UTF-8 or contains a NUL, flagged with an
+    // X-Native-Body-Encoding header that PHPSchemeHandler strips and decodes.
+    //
+    // Keep the PHP below comment-free: it is re-parsed by zend_eval_string on
+    // every request and has to fit eval_code alongside the substituted URI.
+    static char eval_code[16384];
     snprintf(eval_code, sizeof(eval_code),
         "try {\n"
         "    while (ob_get_level() > 0) { ob_end_clean(); }\n"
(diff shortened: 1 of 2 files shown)


simonhamp (approved):
This is fine. Afaict this should only affect the webview, right?

simonhamp:
Thanks for this, @sonalidudhia. Your diagnosis of the response side was right, and it's what started a wider look at the bridge.

That look showed the problem is bigger than binary responses on iOS. Bodies lose bytes in both directions, on both platforms. Request bodies were decoded as UTF-8 and cut at the first NUL. WebKit drops Blob, File and FormData bodies before they reach the scheme handler at all. Android's JS shim turned request bodies into strings, and `NewStringUTF` aborts debuggable builds on binary responses. And `$_FILES` was never filled, so uploads didn't work anywhere. There was also a second iOS response bug this PR doesn't touch: `PHPSchemeHandler` splits on every blank line and keeps only the second piece, which truncates text bodies too.

It's all fixed in one place in #490. That takes a different route from this PR: bodies travel as bytes with an explicit length end to end, rather than being base64-encoded behind a marker header, so there's no header to strip and no size overhead. It covers everything above on both lanes and in classic mode, and it's tested byte for byte on device. Closing this in favour of that one.

(One small correction for the record: #10 is a build failure downloading the PHP binaries, not the Android side of this. The Android equivalent is #431.)

PR #412, opened 2026-09-02
Add runtime appearance overrides without restarting the app

jeremyfrancis55:
## Summary

- add `System::setAppearance()` for `light`, `dark`, and `system` app-local overrides
- add Android and iOS `System.SetAppearance` bridge functions
- drive Android Material chrome and native node dark variants from shared Compose state
- update system bar styling and `System.GetAppearance` without recreating `MainActivity`

## Problem

Using Android's normal day/night delegate recreates the activity so Compose picks up the new theme. In a NativePHP app that also restarts the PHP host and can clear the authenticated runtime session, making an in-app light/dark toggle appear to log the user out. Updating only system bars avoids the restart but leaves Compose/native renderers on the old palette.

This change applies the override as snapshot state instead. Compose recomposes in place, the PID and PHP runtime remain alive, and `AppearanceChanged` still reaches PHP. iOS uses the existing window-level `overrideUserInterfaceStyle` mechanism.

## Testing

- full PHP suite: 928 tests, 3,079 assertions (pass; existing PHP 8.5 deprecation/risky notices remain)
- PHPStan on changed PHP API: pass
- Pint on changed PHP files: pass
- Android Gradle `:app:assembleDebug`: pass in a NativePHP app using these templates
- emulator dark -> light -> dark: full native UI and system bars update; PID remains unchanged and no activity relaunch occurs

## Companion PR

NativePHP/mobile-ui#84 updates every Android plugin renderer to consume the shared state: https://github.com/NativePHP/mobile-ui/pull/84

### DIFF
--- a/resources/androidstudio/app/src/main/java/com/nativephp/mobile/bridge/BridgeFunctionRegistration.kt
+++ b/resources/androidstudio/app/src/main/java/com/nativephp/mobile/bridge/BridgeFunctionRegistration.kt
@@ -29,6 +29,7 @@ fun registerBridgeFunctions(activity: FragmentActivity, context: Context) {
     // plugin). iOS twin: Bridge/Functions/SystemFunctions.swift.
     registry.register("System.OpenAppSettings", SystemFunctions.OpenAppSettings(context))
     registry.register("System.GetAppearance", SystemFunctions.GetAppearance(context))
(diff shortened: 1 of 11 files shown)


jeremyfrancis55:
I realized I can just store the logged in credentials in secure storage..

PR #414, opened 2026-09-02
Surface plugin hook failures instead of building a broken app

### LINKED ISSUE #405: Plugin hook failures are silent: hook output is discarded and a non-zero exit doesn't stop the build
## What happened

My plugin's `copy_assets` hook downloads its prebuilt native libraries on first build. When that download started failing (HTTP 404 — the release it pointed at had been deleted), `php artisan native:run android` still reported success end to end: gradle built, the APK installed and launched — then aborted (SIGABRT) on the first bridge call because the `.so` files the hook should have staged were never there. Nothing in the artisan output or the gradle log mentioned the hook, its error message, or its failing exit code.

## Why (as far as I can tell)

In `src/Plugins/PluginHookRunner.php` (`runPluginHook`):

1. The hook runs via `Artisan::call($command, [...])` — the hook command's own output (`$this->error(...)`, `$this->info(...)`) is captured into Artisan's buffer and never surfaced.
2. A non-zero exit code only produces `$this->warn("Hook {$hookName} ... returned non-zero exit code")` and the build continues.
3. `warn()` / `error()` / `twoColumnDetail()` all no-op when `$this->output` is unset — which appears to be the case in the `native:run android` plugin-compile path, so even the warning and the "Running copy_assets hook" line are invisible there.

Observed on 4.2.0; the code is unchanged in 4.3.0.

## Why it matters

A `copy_assets` hook that stages required native assets is load-bearing: if it fails, the produced app is broken in a way that only shows up as a native crash at runtime, with no build-time trail. That's a rough experience for anyone consuming a plugin — the build says success and the app dies on launch.

## Suggestions

Any of these would have caught it:

- Pass an output buffer/console through to the hook (`Artisan::call($command, $params, $output)`) or print `Artisan::output()` when the hook finishes, so hook errors reach the person building.
- Treat a non-zero hook exit as fatal to the build — or, if some hooks should stay advisory, let `nativephp.json` mark a hook as required.
- At minimum, always emit the non-zero-exit warning to the real console (stderr) rather than the maybe-null `$this->output`.

## Workaround

My plugin now injects its own `preBuild` check into the generated `app/build.gradle.kts` and writes its failure text to a file that check reads, so a failed hook fails gradle with the real cause. Happy to drop that once hook failures surface upstream.

Thanks for NativePHP — the plugin system has otherwise been great to build on.

---

<sub>Investigated and written with AI assistance (Claude), under direction and reviewed by the reporter.</sub>

gwleuverink:
Fixes #405.

A plugin's `copy_assets` hook stages what the plugin needs to work at all. When one fails the build says nothing, finishes green, installs the app, and the app dies on first use. That is what @KevinBatdorf hit: his hook's download started 404ing and `native:run android` reported success end to end.

Reproduced with a throwaway plugin whose `copy_assets` hook prints an error and returns `FAILURE`. The hook ran, failed, and the build printed no hook line, no error and no warning, then `BUILD SUCCESSFUL`, APK installed and launched, exit code 0.

## Why nothing surfaced

The report blames a null `$this->output` in the Android compile path. That part is not it: `RunsAndroid` calls `setOutput($this->output)` and all four `PluginHookRunner` construction sites pass it through.

The real cause is that the runner's output helpers were written against `Command`'s API but are handed a `Command`'s `$this->output`, which is an `Illuminate\Console\OutputStyle`:

- `warn()` guarded on `method_exists($this->output, 'warn')`. `OutputStyle` extends `SymfonyStyle`, which has `warning()` and no `warn()`, so the non-zero-exit warning never printed on any path.
- `twoColumnDetail()` guarded on `isset($this->output->components)`. `$components` lives on the Command and is `protected`, so that is false for every object the build passes. The "Running <hook> hook" line has never printed either.

Same guard style as the compiler `warn()` methods fixed in #410, which is where I took the fallback pattern from.

Three things had to line up for the build to stay green, so all three are addressed below: the hook's own output went to a `BufferedOutput` nobody read, a non-zero exit only warned through the dead `warn()`, and `runAndroid()`/`runIos()` were `void` under an unconditional `return self::SUCCESS`.

## Changes

**Surface hook warnings the build was dropping.** The four helpers now try the `Command` method, then fall back to `writeln()`, matching #410. `twoColumnDetail()` falls back to the same `View\Components\Factory` the Command builds it with. This also un-silences two warnings that were never reaching anyone: a manifest `assets` entry whose source file is missing, and an unresolved `${ENV_VAR}` during placeholder substitution.

**Stream hook output and stop the build when a hook fails.** `Artisan::call` now receives the output, so a hook explaining itself is heard. A non-zero exit throws `PluginHookFailedException`, which the compile wrappers turn into a `false` and the build stops before Gradle. Those wrappers now catch `\Throwable` rather than `\Exception`, since a hook that fatals throws an `Error`.

**Let a failed build reach `native:run`'s exit code.** `runAndroid()`, `runIos()`, `runTheIosBuild()`, `runOnSimulator()` and `runOnRealDevice()` return a status and `handle()` carries it out. Every early return in all of them was already a failure path; a launch that fails after a successful install stays a warning, since the app is there to tap. This is what lets CI notice, and it also covers the Gradle-failed and xcodebuild-failed cases, where `native:run` previously exited 0 too. Adjacent to #76, which does the same for `native:package`; happy to drop this commit if you would rather it land there.

**Keep a failed `post_build` hook from unwinding a finished build.** Those hooks run after the app is built and installed, so a failure has nothing left to stop. Both call sites catch and warn instead of letting the exception escape, which would otherwise skip hot reload on Android and the install and launch on iOS. The compile-time wrappers also print the underlying file and line when a hook fatals, so a plugin author gets more than a message.

**Tests.** There were none for hook running. Five now cover the announce line, hook output reaching the console, a non-zero exit failing the build, a missing hook command, and a plugin without the hook being left alone. They run the runner against a real `OutputStyle`, which is the shape that was broken.

## After

```
 Compiling plugin .. probe/hook-405 (dev-main)
 Running copy_assets hook .. probe/hook-405
HOOK405 ERROR: simulated 404 downloading prebuilt .so files
❌ Plugin compilation failed: Hook copy_assets for probe/hook-405 failed: exit code 1
```

Build stops before Gradle, exit code 1. A passing hook still exits 0, builds, installs and launches.

## Verified

On device, both platforms, both ways, using a plugin whose `copy_assets` hook can be flipped between success and failure:

| | Android emulator | iOS simulator |
|---|---|---|
| failing hook | exit 1, stops before Gradle | exit 1, stops before xcodebuild |
| passing hook | exit 0, APK installed and launched | exit 0, app installed and running |

`native:package android --build-type=bundle` behaves the same: a failing hook stops it before Gradle with the error shown, a passing one produces the signed bundle. It still exits 0 either way, because `PackageCommand::handle()` is `void`; that is what #76 fixes and I left it alone.

Other commands are unaffected. `native:watch` and `native:release` never reach plugin compilation or the hook runner, and `native:install` does not either. `native:build` is exercised by the iOS run path.

- `--watch` on Android still builds, launches and enters hot reload.
- Full suite: 925 passed. Pint clean. PHPStan reports the same 14 pre-existing errors as `main` untouched.

### DIFF
--- a/src/Commands/BuildIosAppCommand.php
+++ b/src/Commands/BuildIosAppCommand.php
@@ -1213,6 +1213,12 @@ class BuildIosAppCommand extends Command
         } catch (\Throwable $e) {
             $this->error("❌ Plugin compilation failed: {$e->getMessage()}");
 
+            // A hook that fataled carries the real error underneath. Without
+            // this the author sees the message and no idea where it came from.
+            if ($cause = $e->getPrevious()) {
+                $this->line("   at {$cause->getFile()}:{$cause->getLine()}");
+            }
+
             return false;
         }
     }
@@ -1241,7 +1247,14 @@ class BuildIosAppCommand extends Command
             output: $this->output
         );
 
-        $hookRunner->runPostBuildHooks();
+        // The app is already built by now, so a post_build hook that fails has
+        // nothing left to stop. Say so and leave the run green rather than
+        // unwinding over a finished build.
+        try {
+            $hookRunner->runPostBuildHooks();
+        } catch (\Throwable $e) {
+            $this->warn("⚠️  {$e->getMessage()}");
+        }
     }
 
     /**
--- a/src/Commands/BuildIosAppCommand.php
+++ b/src/Commands/BuildIosAppCommand.php
@@ -1210,7 +1210,7 @@ class BuildIosAppCommand extends Command
             $compiler->compile();
 
             return true;
-        } catch (\Exception $e) {
+        } catch (\Throwable $e) {
             $this->error("❌ Plugin compilation failed: {$e->getMessage()}");
 
             return false;
--- a/src/Commands/BuildIosAppCommand.php
+++ b/src/Commands/BuildIosAppCommand.php
@@ -1247,14 +1247,7 @@ class BuildIosAppCommand extends Command
             output: $this->output
         );
 
-        // The app is already built by now, so a post_build hook that fails has
-        // nothing left to stop. Say so and leave the run green rather than
-        // unwinding over a finished build.
-        try {
-            $hookRunner->runPostBuildHooks();
-        } catch (\Throwable $e) {
-            $this->warn("⚠️  {$e->getMessage()}");
-        }
+        $hookRunner->runPostBuildHooks();
     }
 
     /**
--- a/src/Commands/PackageCommand.php
+++ b/src/Commands/PackageCommand.php
@@ -63,7 +63,7 @@ class PackageCommand extends Command
 
     protected string $platform;
 
-    public function handle(): void
+    public function handle(): int
     {
         // Get platform (flags take priority over argument)
         if ($this->option('ios')) {
@@ -75,7 +75,7 @@ class PackageCommand extends Command
             if (! $platform) {
                 \Laravel\Prompts\error('Platform must be specified via argument or flags (--ios/--android)');
 
-                return;
+                return self::FAILURE;
             }
             // Support shorthands: 'a' for android, 'i' for ios
             $this->platform = match (strtolower($platform)) {
@@ -88,7 +88,7 @@ class PackageCommand extends Command
         if (! in_array($this->platform, ['android', 'ios'])) {
             \Laravel\Prompts\error('Platform must be either "android" or "ios" (or "a" / "i" as shortcuts)');
 
-            return;
+            return self::FAILURE;
         }
 
         $this->validateAppId();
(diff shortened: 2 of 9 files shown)


shanerbaner82 (commented):
Ran this against the demo app with mobile-ui's font hook passing and then forced to fail. Both behave as described: exit 0 with the hook line printing for the first time, exit 1 before Gradle on the second. Nice.

Three things I'd like fixed before it goes in, all small:

1. A failed post_build hook now skips every later plugin's post_build hook. `runHook()` throws out of its foreach, and both call sites catch outside the loop, so plugin B never runs when plugin A's hook exits 1, and nothing says so. Before this PR the loop warned and carried on. The catch wants to be per plugin inside `runPostBuildHooks`.

2. `runOnSimulator()` still returns true no matter what. None of its three task closures look at the Process result, so a failed `simctl install` prints "App launched!" and exits 0. The real-device path tracks `$installFailed`; the simulator path needs the same.

3. `native:package` still exits 0 on a failed hook. You deferred that to #76, but #76 got closed unmerged today. Giving `PackageCommand::handle()` an int return and returning FAILURE on the early returns, same as you did for `RunCommand`, closes it here.

The rest (opposite exit codes on iOS vs Android for "installed but launch failed", the vendor file:line hint, the output helper ladders) can wait.

gwleuverink:
Thanks, all three fixed and pushed. Rebased on main too.

1. Moved the catch into `runPostBuildHooks()` so it wraps each plugin on its own. Tested with two probe plugins on both platforms: A's hook fails, you get the warning, B's hook still runs, app still installs and launches.

2. `runOnSimulator()` now checks the install and launch results like the device path already did. I made a hook delete `Info.plist` from the built app to force it: install fails, you get "App installation failed!" and exit 1

3. `native:package` returns a proper exit code now. On Android a failing hook gives exit 1 before Gradle, a passing one gives 0 with a signed bundle. On iOS the failing hook also exits 1, and so does a failed archive

PR #415, opened 2026-09-03
Mark non-production builds as testing builds for both stores

simonhamp:
## What

A build made in any environment other than `production` is a testing build. Both stores are now told so **at package time**, so the declaration travels with the artifact and a testing build cannot later be promoted to a public release by hand.

Three pieces, one per store surface:

**Google Play — manifest audience.** `AndroidManifest.xml` gains a `com.google.android.play.largest_release_audience` meta-data entry set to `CLOSED_TESTING`. Play holds the artifact to that ceiling. The native project is reused between builds, so the previous declaration is always stripped first — moving back to `production` does not leave a stale ceiling behind.

**App Store Connect — TestFlight.** An `app-store` export sets `testFlightInternalTestingOnly` in the export options, confining the upload to TestFlight's internal testers. Without it, the same archive could be handed to external testers or submitted for review. Other export methods (`ad-hoc`, `enterprise`, `debugging`) are unaffected, where the key has no meaning.

**Play Store upload — refuse the production track.** A non-production build's manifest declares a closed testing audience, so Play would reject the production track anyway. The upload is now refused up front, where the reason can still be explained, instead of failing later against the API. Testing tracks (`internal`, etc.) are still reachable.

The shared predicate lives in a new `DeclaresReleaseAudience` trait, so "is this a production build?" is answered in one place across iOS packaging, Android preparation, and Play publishing.

## Why

Nothing previously distinguished a build made with `APP_ENV=staging` from a real release once the artifact existed. The environment was a property of the machine that ran the build, not of the thing it produced, so a testing build could be uploaded to a production track or promoted through App Store Connect with no signal that it was never meant to ship. Marking the artifact itself moves that check to where it cannot be skipped.

## Testing

- `composer test` — 933 passed (3099 assertions), 3 skipped, 4 risky.
- `composer format` — pass.
- New `tests/Feature/ReleaseAudienceTest.php` covers the manifest: the entry is written for testing builds, sits inside `<application>`, is absent for production, is stripped when moving back to production, is written only once across repeated builds, and a missing manifest is left alone.
- New `tests/Feature/TestingBuildStoreGuardsTest.php` covers both store-side halves: the iOS export options with and without the restriction (and that it is confined to `app-store` exports), and that the Play upload refuses `production` from a testing build while still allowing testing tracks. It uses `Http::preventStrayRequests()` so the guard is proven to turn the upload away before any API work begins.
- Existing test doubles that compose `PreparesBuild` gained the new `updateReleaseAudience()` stub.

## Note

`composer analyse` reports 3 pre-existing PHPStan errors in `src/Concerns/RunsAndroid.php` (lines 97/113/132, about `native:package` options `watch`/`udid`). Verified against an unmodified worktree at the base commit — they are unrelated to this change and are not introduced here. Also note PHPStan needs `--memory-limit=2G`; the default 128M crashes the parallel worker.

## Summary by CodeRabbit

- **New Features**
  - Non-production Android builds are labeled as non-production releases.
  - Non-production iOS App Store exports are restricted to internal TestFlight testing.
  - Production publishing to Google Play requires a production build.

- **Bug Fixes**
  - Prevents testing builds from being distributed through production release channels.
  - Preserves developer-defined Android release audience restrictions.

- **Tests**
  - Added coverage for Android release labeling and iOS and Google Play distribution safeguards.

### DIFF
--- a/src/Concerns/DeclaresReleaseAudience.php
+++ b/src/Concerns/DeclaresReleaseAudience.php
@@ -1,0 +1,42 @@
+<?php
+
+namespace Native\Mobile\Concerns;
+
+/**
+ * A build made in any environment other than production is a testing build,
+ * and both stores are told so at package time: Google Play through the
+ * manifest's largest release audience, App Store Connect through TestFlight
+ * internal testing. That declaration travels with the artifact, so a testing
+ * build cannot later be promoted to a public release by hand.
+ */
+trait DeclaresReleaseAudience
+{
+    /**
+     * The AndroidManifest meta-data Google Play reads to learn how far a
+     * release may travel.
+     */
+    protected const LARGEST_RELEASE_AUDIENCE_KEY = 'com.google.android.play.largest_release_audience';
+
+    protected function buildsForProduction(): bool
+    {
+        return config('app.env') === 'production';
+    }
+
+    /**
+     * The largest audience this build may be released to, or null when it is
+     * a production build and no ceiling applies.
+     */
+    protected function largestReleaseAudience(): ?string
+    {
+        return $this->buildsForProduction() ? null : 'CLOSED_TESTING';
+    }
+
+    /**
+     * Whether an App Store Connect upload from this build must be confined to
+     * TestFlight's internal testers.
+     */
+    protected function restrictToInternalTestFlight(): bool
+    {
+        return ! $this->buildsForProduction();
+    }
+}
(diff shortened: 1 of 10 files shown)


SRWieZ:
Could I still promote a TestFlight build to production ? 

I'm not sure to understand the change in practical terms.

simonhamp:
@SRWieZ when building with anything other than `production` as your `APP_ENV` value, you probably don't want to be able to promote that build to "Production Release"

This enforces that at a code level to make sure the stores don't allow you to release it.

shanerbaner82 (approved):

Issue #308, opened 2026-08-10
[Masterclass] iOS: Tapping outside a text field does not dismiss the keyboard on screens with native chrome


shrutibalasawebdev:
### What were you trying to do?

Dismissing the keyboard by tapping outside a text field, on a screen (chat conversation screen) that uses native chrome (a tab bar / stack navigation), rather than on a plain chrome-less screen.

### What happened?

Expected: tapping anywhere outside the focused text field dismisses the keyboard, consistently across all screens.

Actual on iOS: on a screen with native chrome, tapping outside the field does nothing - the keyboard stays up. The only way to dismiss it is to drag the list downward, which forces it away.

On a plain screen with no native chrome (e.g. a login screen), tapping anywhere outside the field dismisses the keyboard as expected. So the behaviour differs between the two screen types within the same app.

### How to reproduce the bug

1. Build a screen with no native chrome containing a text input. Focus the field, then tap elsewhere on the screen - the keyboard dismisses.
2. Build a screen inside a tab or stack layout containing a text input and a list. Focus the field, then tap elsewhere - the keyboard stays up.
3. Drag the list downward on that second screen - the keyboard is dismissed.

Reproduced on iOS.

### Debug Output

```

   INFO  NativePHP Mobile.  

+--------------------+--------+
| Package Version    | 4.1.0  |
| PHP Version (Host) | 8.4.18 |
| OS                 | Darwin |
| OS Version         | 25.5.0 |
| Embedded PHP       | 8.4.24 |
+--------------------+--------+

   INFO  Installed Plugins.  

+--------------------------+---------+
| Package                  | Version |
+--------------------------+---------+
| nativephp/mobile-browser | 1.0.2   |
| nativephp/mobile-ui      | 0.3.0   |
+--------------------------+---------+

   INFO  Development Tools.  

+----------------+--------------------------------------+
| Xcode          | Xcode 26.5                           |
| Android Studio | Android Studio AndroidStudio2025.2.3 |
| Gradle         | Gradle 8.14.5                        |
| Java           | openjdk version "17.0.18" 2026-01-20 |
| CocoaPods      | CocoaPods 1.16.2                     |
+----------------+--------------------------------------+
```

### Which operating systems have you seen this occur on?

macOS

### Which platforms were you trying to build for?

iOS (Device)

### Notes

[...]

PR #423, opened 2026-09-08
fix(ios): let the window background reach tab-hosted screens

techenby:
## The problem

`UI.SetBackground` writes `WindowBackgroundState.shared.color`. Three surfaces read it:

| | |
|---|---|
| `ContentView.swift:22` | the base layer every native screen renders over |
| `NativeRootStackRenderer.swift:275` | per stack screen, via `StackScreenBackgroundModifier` |
| `SetBackground.execute` | the UIKit windows |

`NativeRootTabsRenderer` does not — zero references. TabView hosts its content on its own
`systemBackground` container, which covers the base layer `ContentView` paints underneath.

So on an app whose routes are all under a `usesNativeChrome()` layout with a `tabBar()`,
`UI.SetBackground` has **no observable effect anywhere**. A dark-themed app renders on a
white page ground on a light-appearance device, whatever the `background` token says.

This is the same defect `StackScreenBackgroundModifier` was written for — its own doc comment
describes the symptom:

> NavigationStack draws its own `systemBackground` container behind screen content with no
> SwiftUI override hook — without this, a dark app shows a white band in the bottom safe-area
> inset on every stack screen.

The tabs twin was never written.

## The fix

`TabScreenBackgroundModifier`, mirroring the stack twin, attached to the screen content inside
`PerTabContent` rather than to the TabView — same reasoning as the neighbouring
`dismissesKeyboardOnTap`, which is deliberately kept off the TabView to avoid iOS 26's search
capsule (#308). It's a no-op when no override is set, so stock appearance is untouched.

Follows the file-local twin convention already used for
`StackBottomBarInsetModifier`/`BottomBarInsetModifier` and
`StackBarBackgroundModifier`/`ExplicitBarBackgroundModifier`.

## Reproducing

1. A layout with `usesNativeChrome() == true` and a `tabBar()`, attached via `Route::nativeGroup()`.
2. A dark `background` token in `config/native-ui.php` (e.g. `#191722`).
3. `nativephp_call('UI.SetBackground', json_encode(['color' => theme('background')]))` at boot.
4. Run on a device set to **light** appearance.

Expected the page ground to be `#191722`; it is white.

A confirming variant, before this patch: add `->dark()` to the `TabBar`. The ground becomes
**pure black** rather than `#191722` — `preferredColorScheme(.dark)` changes what
`systemBackground` resolves to, which shows it is `systemBackground` being painted and not the
window color.

Found on `nativephp/mobile` 4.3.2, iPhone 17 Pro simulator / iOS 26.5.

## Verification

Built and run on an iPhone 17 Pro simulator (iOS 26.5) against a tabs-only app with
`background` = `#191722`.

Sampling the rendered framebuffer, **before** this patch (with `->dark()` on the `TabBar`, so
`systemBackground` resolves dark): `#000000` at every background sample point — the window color
is absent. **After**: `#191722` at every point, including both the status-bar strip and the
region below the tab bar.

```
  #191722   status-bar strip (top inset)
  #191722   page ground (upper / mid / lower)
  #191722   below the tab bar (bottom inset)
```

`** BUILD SUCCEEDED **`, no new warnings.

## Two adjacent gaps, not fixed here

Mentioning these since they're the same story, but deliberately out of scope for this PR:

1. **Nothing ever paints the window from the theme.** The framework knows the `background` token
   and the docs call it "the page root behind everything", but no code path pushes it to the
   window. The only caller of `UI.SetBackground` in the framework is
   `Edge/BenchmarkComponent.php:378`, with a hardcoded `#0F172A`, and there's no facade or docs
   page — so every dark-themed app hits this and has to find an undocumented bridge call.
   Painting from `theme('background')` at boot and on `AppearanceChanged` would make it correct
   by default.

2. **The published starter layout teaches a pattern that can't work under chrome.**
   `mobile-ui`'s `stubs/views/components/layouts/app.blade.php` puts `bg-theme-background` on the
   screen root (lines 43 and 49). Under a chrome layout that fill is laid out in the inset box, so
   it stops short of the screen edge while scroll content keeps drawing past it under the Liquid
   Glass bar — the boundary reads as a hard line above the tab bar. Different repo; happy to open
   that separately.

https://claude.ai/code/session_01RRCNbc5FrcTjiD8aXUfKTA

### DIFF
--- a/resources/xcode/NativePHP/NativeRender/NativeRootTabsRenderer.swift
+++ b/resources/xcode/NativePHP/NativeRender/NativeRootTabsRenderer.swift
@@ -569,6 +569,12 @@ private struct PerTabContent: View {
                 // business, and wrapping the TabView risks iOS 26's search
                 // capsule (mobile-air #308).
                 .dismissesKeyboardOnTap()
+                // TabView hosts its content on its own `systemBackground`
+                // container, which covers the base layer ContentView paints —
+                // so without this the window background never reaches a tabbed
+                // screen. Attached here rather than to the TabView for the same
+                // reason as `dismissesKeyboardOnTap` above (#308).
+                .modifier(TabScreenBackgroundModifier())
         } else {
             Color.clear
         }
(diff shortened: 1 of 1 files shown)


unlocdavid:
I have opened a PR before with a similar fix and extra fixes for some tab bar and keyboard rendering weirdness: #358 

Just an FYI and probably best to review both PR's together.

techenby:
Then I probably should close this in favor of that right?

simonhamp:
Closing in favor of #358

PR #435, opened 2026-09-11
Add native:screenshot command for capturing device/simulator screenshots

lessevv:
## Summary
- Adds `native:screenshot {os} {udid?} --output=`, capturing a PNG from a booted iOS simulator (`xcrun simctl io`) or connected Android device/emulator (`adb exec-out screencap`)
- Adds `--crop=top|bottom|both` with `--crop-percent` and `--crop-offset` to trim OS chrome (status bar, home indicator/nav bar) or keep just a strip near one edge, so docs/marketing screenshots don't need manual cropping
- iOS target auto-selects the booted simulator when unambiguous instead of always prompting
- Android target auto-resolves when exactly one device is connected, with clear errors for zero or multiple devices

## Test plan
- [x] `vendor/bin/pest --filter=Screenshot` passes (27 tests)
- [x] Full suite passes locally

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,6 +56,8 @@ claude.md
 test-keystore.jks
 .secrets
 .phpunit.cache/test-results
+/tests/coverage
+.phpactor.json
 .github/workflows/test-package-local.yml
 
 /composer.lock
--- a/src/Commands/ScreenshotCommand.php
+++ b/src/Commands/ScreenshotCommand.php
@@ -1,0 +1,190 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Native\Mobile\Commands;
+
+use Illuminate\Console\Command;
+use Illuminate\Support\Facades\Process;
+use Native\Mobile\Support\HostOperatingSystem;
+
+final class ScreenshotCommand extends Command
+{
+    /** `xcrun simctl`'s own alias for "whichever simulator is currently booted". */
+    private const string DEFAULT_IOS_TARGET = 'booted';
+
+    protected $signature = 'native:screenshot
+        {os : Platform (android/a or ios/i)}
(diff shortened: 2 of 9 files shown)


shanerbaner82:
tldr: two P2 issues to fix before merging:

- **[P2] Numeric Android serials crash auto-selection** — [ScreenshotCommand.php:294](https://github.com/NativePHP/mobile-air/blob/efb60961ea9ef2cad6218a3ea267018b879d0812/src/Commands/ScreenshotCommand.php#L294). PHP converts numeric array keys to integers, so `array_key_first()` violates the `?string` return type. Reproduced with an ADB fixture containing serial `1234567890`: auto-selection throws a `TypeError` before capture. Return the stored string value or cast the key, and cover numeric serials in a test.

- **[P2] iOS auto-selection can choose a non-iOS simulator** — [RunsIos.php:429](https://github.com/NativePHP/mobile-air/blob/efb60961ea9ef2cad6218a3ea267018b879d0812/src/Concerns/RunsIos.php#L429). The code flattens every runtime returned by `simctl`, including watchOS/tvOS. With only one such simulator booted, `native:run ios` selects it without prompting. Reproduced with a watchOS fixture: the resolver returned the watch UDID despite being given an iPhone target list. Filter to supported iOS targets before auto-selecting, and test this case.

Validation: 27 screenshot tests passed using existing local dependencies. Four iOS tests were blocked by local Testbench cache permissions. No device testing performed.

lessevv:
Thanks for the thorough review — both are real bugs, fixed in a6b9f93:

- `resolveAndroidTarget` now returns `reset($devices)` (the value) instead of `array_key_first()` (the key), since a purely-numeric serial gets cast to an int key by PHP. Added a regression test with serial `1234567890`.
- `getBootedIosSimulatorUdids` now filters to `SimRuntime.iOS-*` runtime keys before flattening, so a lone booted watchOS/tvOS simulator can't be auto-selected. Added a regression test with only a watchOS simulator booted.

PR #466, opened 2026-09-16
fix(ios): boot and open simulators for iOS

### LINKED ISSUE #458: native:run ios fails to open the simulator UI on Xcode 27 (Simulator.app replaced by DeviceHub.app)
### Description

Xcode 27 removes the standalone `Simulator.app` and replaces it with **Device Hub**
(`DeviceHub.app`, bundle id `com.apple.dt.Devices`). `RunsIos.php` still shells out to
`open -a Simulator`, which no longer resolves:

```
Unable to find application named 'Simulator'
```

The developer tools were also relocated in Xcode 27: `Contents/Developer/Applications/`
no longer exists, and the bundled apps now live in `Contents/Applications/`.
`SimulatorKit.framework` moved from `Contents/Developer/Library/PrivateFrameworks/` to
`Contents/SharedFrameworks/`.

### Impact

Non-fatal, but confusing. The call at `RunsIos.php:255` is a bare `shell_exec()` whose
result is never checked, so the run continues: `xcrun simctl boot / install / launch` all
work normally under Xcode 27 and the app *is* deployed and started. Only the simulator
window never opens, so the user sees an error and an apparently dead run — while the app
is actually running headless on a booted device.

### Environment

| | |
|---|---|
| macOS | 27.0 (26A428) |
| Xcode | 27.0 (27A266a) |
| `nativephp/mobile` | 4.4.1 |
| `nativephp/mobile-ui` | 0.3.0 |
| PHP | 8.4.23 |
| Laravel | 13.31.0 |

### Steps to reproduce

1. Install Xcode 27.
2. Run `php artisan native:run ios` against a simulator target.
3. Observe `Unable to find application named 'Simulator'`; the build completes but no
   simulator window appears.

### Root cause

`vendor/nativephp/mobile/src/Concerns/RunsIos.php:255`

```php
shell_exec('open -a Simulator');
```

### Suggested fix

Resolve the UI app from the active developer directory instead of relying on a hardcoded
app name, so both Xcode ≤ 26 and Xcode 27+ are covered:

```php
$developer = rtrim(trim((string) shell_exec('xcode-select -p')), '/');

$simulatorUi = collect([
    $developer.'/Applications/Simulator.app',           // Xcode <= 26
    dirname($developer).'/Applications/DeviceHub.app',  // Xcode 27+
])->first(fn (string $path): bool => is_dir($path));

if ($simulatorUi !== null) {
    shell_exec('open '.escapeshellarg($simulatorUi));
}
```

This also removes the dependency on LaunchServices having the app indexed, which is a
separate long-standing source of `open -a` flakiness.

### References

Same breakage reported in other toolchains:

- expo/expo#46882 — `expo run:ios` fails because `Simulator.app` was replaced by `DeviceHub.app`
- software-mansion/argent#406 — `boot-device` fails on Xcode 27
- anthropics/claude-code#79991 — iOS Simulator breaks under Xcode 27 (Device Hub)

### What happened?

see up there

### How to reproduce the bug

see up there

### Debug Output

...

macOS

iOS (Simulator)

### Notes

_No response_

vikas5914:
Fixes #458.

- `native:emulator ios` routed to `startAndroid()` — new `LaunchesIosSimulator` (simctl JSON list/boot, interactive pick), `ios` branch now calls `startIos()`.
- Xcode 27 removed `Simulator.app` (now `DeviceHub.app` under `Contents/Applications`): resolve the UI by path, not `open -a <name>` — Simulator.app (≤26) first, DeviceHub via `devices://manage/select?id=<udid>` second, headless warning otherwise. `-CurrentDeviceUDID` only for freshly booted devices.
- `RunsIos::runOnSimulator` uses the same helper instead of bare `shell_exec('open -a Simulator')`.


keiserjb:
Yep, I ran into this issue as well.

simonhamp:
Yep, me too. Looks like this nails it

simonhamp:
Thanks @vikas5914

PR #468, opened 2026-09-17
Remove timeout on robocopy process

adamjmarshall:
## Summary
- `BundleFileManager::copyWithRobocopy()` ran robocopy through Laravel's `Process::run()`, which defaults to a 60-second timeout
- Large app bundles can take longer than that to copy, causing the copy to be killed mid-flight with a timeout exception
- Changed to `Process::timeout(0)->run(...)` so the robocopy process runs to completion regardless of size

## Test plan
- [x] Existing `AndroidBundleCopyTest` suite passes unchanged (uses `Process::fake()`, unaffected by the timeout change)

?? Generated with [Claude Code](https://claude.com/claude-code)


src/Support/BundleFileManager.php
```diff
@@ -132,7 +132,7 @@ private static function copyWithRobocopy(string $source, string $destination, ar
         $sourceWin = str_replace('/', '\\', $source);
         $destinationWin = str_replace('/', '\\', $destination);
 
-        $result = Process::run("robocopy \"{$sourceWin}\" \"{$destinationWin}\" /MIR /NFL /NDL /NJH /NJS /NP /R:0 /W:0{$excludeArgs}");
+        $result = Process::timeout(0)->run("robocopy \"{$sourceWin}\" \"{$destinationWin}\" /MIR /NFL /NDL /NJH /NJS /NP /R:0 /W:0{$excludeArgs}");
```
simonhamp: Why not [`forever()`](https://laravel.com/framework/docs/processes#:~:text=may%20invoke%20the-,forever,-method%3A)


```suggestion
        $result = Process::forever()->run("robocopy \"{$sourceWin}\" \"{$destinationWin}\" /MIR /NFL /NDL /NJH /NJS /NP /R:0 /W:0{$excludeArgs}");
```

PR #478, opened 2026-09-19
Set NATIVEPHP_RUNNING before any artisan runs, and migrate once per platform

simonhamp:
On Android, apps upgraded to 4.5.0 open the web view instead of their `routes/mobile.php` screens. Logcat says `Boot plan for /: WEB_LEGACY (0 native patterns)`. A user reported it the day after release.

## Cause

The Android shell runs `config:cache` on the first boot after an install or update. `NATIVEPHP_RUNNING` isn't set at that point. Only the PHP runtime sets it, later. So the cache is written with `nativephp-internal.running => false` and every boot after that reads the cached false.

4.5.0 only loads `routes/mobile.php` when `running` is true (#474), so native routes never register on the device. The route manifest the shell reads comes out empty and it falls back to the web view.

The stale value is older than 4.5.0. `config:cache` came in with #173, and since then the temp and `mobile_public` disks and the `android-hot` Vite file have also been switched off on Android.

## Fix

Both shells now set `NATIVEPHP_RUNNING` in `setupEnvironment()`, before any PHP runs. On Android that means `config:cache` records `running => true`.

With the flag on, `migrate` also loads the package's queue migrations. They're renamed to `9999_12_31_000000_create_jobs_table` and `9999_12_31_000001_reshape_failed_jobs_table`, so they sort after every app migration. The app's own queue tables always get created first, including dated ones like Laravel 10's `2019_08_19_000000_create_failed_jobs_table`, and ours only add what's missing. #479 gave them `000003`/`000004` names, which still sorted before dated migrations. Those names never shipped in a release.

Because of that, each platform now migrates once per install or update:

- Android keeps its single `migrate` in `runBaseArtisanCommands()`, now with the flag on.
- iOS used to migrate twice: once in `AppUpdateManager` without the flag, then again in the persistent runtime with it. The persistent one is gone. The `AppUpdateManager` one is the only one both runtime modes reach, so classic mode never ran the package migrations before this. `migrateDatabase()` in `NativePHPApp.swift` has no callers and has been dead since the first commit. I left it alone.

## What to watch for

Existing Android apps only get this once their Android project is refreshed with `php artisan native:install android --force`. `composer update` doesn't do it on its own, because `setupComposerPostUpdateScript()` has been disabled since the first commit. This should go in the 4.5.1 release notes. iOS apps on an older shell keep working, because their old double migrate still runs ours last.

Existing installs keep a leftover `0001_01_01_000001_create_jobs_table` row in `migrations`. It does no harm.

A Laravel 10 app has no `job_batches` table, so our create migration adds it. It leaves `jobs` and `failed_jobs` alone when the app already has them.

Now that `running` is true on Android, `registerFilesystems()` adds the `temp` and `mobile_public` disks. It will replace an app's own `temp` disk, same as on iOS. `configureViteHotFile()` now reads `public/android-hot`, which is the file the Vite plugin writes for Android.

## Testing

Suite: 1128 passed, 3 skipped, 4 risky. `JobsMigrationsTest` has a new case: Laravel 10's dated `failed_jobs` and `queue:table` migrations run in the same migrate as ours with no clash. Renaming ours back to `0001_01_01_000003` makes it fail with "table failed_jobs already exists".

On an API 35 emulator and an iPhone 16 Pro simulator, with the starter app plus an extra dated app migration:

- Fresh install, both platforms: one migrate, and every migration lands in batch 1 with ours last. On Android the boot plan is `NATIVE_DIRECT (1 native patterns)`, the cached config has `running => true`, and the native home screen shows.
- App with Laravel 10-style dated queue migrations, both platforms: one migrate, no errors, ours last.
- Upgrading over the previous build of this branch without uninstalling, both platforms: the `9999_*` migrations run once as no-ops.
- iOS classic mode: one migrate, ours included.
- `main`'s Android shell reproduces the bug: `WEB_LEGACY (0 native patterns)`, cached `running => false`, a 404.


shanerbaner82 (approved):
Confirmed independently against tag `4.5.0`:

**Root cause**
- `LaravelEnvironment.setupEnvironment()` sets `NATIVEPHP_PLATFORM=android` but never `NATIVEPHP_RUNNING`
- `runBaseArtisanCommands()` then runs `config:cache` (present since #173 / pre-4.5)
- `config/nativephp-internal.php` defaults `'running' => env('NATIVEPHP_RUNNING', false)` → cache bakes `false`
- `#474` (4.5.0) gated `routes/mobile.php` on `shouldLoadMobileRoutes()` → `config('nativephp-internal.running')`, so fresh Android installs get `WEB_LEGACY (0 native patterns)`

**iOS:** not affected — post-extraction artisan is only `migrate --force` + `storage:link`; no `config:cache`. `NATIVEPHP_RUNNING` is also set earlier in the iOS boot path.

**This PR's approach looks right.** Double-migrate (app first, then flag on for package migrations + caches) matches the #479 ordering constraint. Avoid the `running` default-from-platform shortcut in `nativephp-internal.php` upstream — that would load package migrations on the first migrate and reintroduce the dated `jobs` clash.

**Release note reminder:** existing apps need `php artisan native:install android --force` (composer update alone won't refresh the Kotlin shell). App-side workaround until then: override `config/nativephp-internal.php` `'running'` as in the user report, then reinstall so `config:cache` rewrites.

One optional follow-up (not blocking): restoring the short-lived `if (env('NATIVEPHP_RUNNING')) { config(['nativephp-internal.running' => true]); }` after `mergeConfigFrom` would heal already-poisoned installs on `composer update` alone, for users who skip `native:install --force`. Happy to open that as a tiny follow-up if you want both belts.

Emulator matrix in the PR description is convincing. LGTM.

